PR #353, opened 2020-10-02
Version shared state at 0 when event nil if never shared state otherwise latest

nporter-adbe:
When the Event is nil shared state will be versioned at 0 if this extension is yet to publish a shared state, otherwise, it will be published at the latest shared state version


shalehaha:
hmmmm.. how could code cov changed so much

shalehaha (approved):

nporter-adbe:
@sbenedicadb any feedback?

AEPCore/Sources/eventhub/EventHub.swift
```diff
@@ -261,13 +261,18 @@ final class EventHub {
             return nil
         }
 
-        var version = 0 // default to version 0
+        guard let sharedState = extensionContainer.sharedState else { return nil }
+
+        // default to version 0
+        var version = 0
         // attempt to version at the event
         if let unwrappedEvent = event, let eventNumber = eventNumberMap[unwrappedEvent.id] {
```
sbenedicadb, line 269: we're supposed to shadow the var name when unwrapping according to our style guide, right?

sbenedicadb (approved):
looks good.  i'd prefer some of the longer doc lines explaining parameters to go into the discussion of the method, but that's my preference.  pending your update from main, this is ready to merge.

PR #358, opened 2020-10-05
Getting started doc

nporter-adbe:
Getting started doc.

Anything I missed?

### DIFF
--- a/Documentation/GettingStarted.md
+++ b/Documentation/GettingStarted.md
@@ -1,8 +1,8 @@
 # Getting Started
 
-This guide will walk you through how to get up and running with the AEP Swift SDK with only a few lines of code.
+This guide walks through how to get up and running with the AEP Swift SDK with only a few lines of code.
 
-> If you are an existing ACP SDK customer, check out the [migration document](./Migration/ACP-Migration.md).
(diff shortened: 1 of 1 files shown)


Documentation/GettingStarted.md
```diff
@@ -23,16 +27,77 @@ import AEPSignal
 2. Register the core extensions, configure the core with your assigned application identifier, and start lifecycle metrics.
 To do this, you add the following code to your Application Delegate's `application(_:didFinishLaunchingWithOptions:)` method:
```
shalehaha, line 28: we can remove lifecycle related words and code from this step, since now we have a new segment for it

Documentation/GettingStarted.md
```diff
@@ -23,16 +27,77 @@ import AEPSignal
...
+}
 ```
 
+When the app enters the background, pause Lifecycle data collection from your app's `applicationDidEnterBackground:` delegate method:
+
+```swift
+func applicationDidEnterBackground(_ application: UIApplication) {   
+  AEPCore.lifecyclePause()
+}
+```
```
shalehaha, line 90: should also add the sample code if the app is using App Scene. Don't think we need to do it now, but later we will also need sample code for the new SwiftUI lifecycle.

sbenedicadb (approved):
looks great

PR #360, opened 2020-10-06
Add steps on including assurance

nporter-adbe:
Adds a debugging doc to get started with assurance.

### DIFF
--- a/Documentation/Debugging.md
+++ b/Documentation/Debugging.md
@@ -1,6 +1,6 @@
 # Implementation Validation with Assurance
 
-The AEP SDK offers an extension to inspect quickly, validate, debug data collection, and experiences for any mobile app using the AEP SDK. We built Assurance to do the heavy lifting of getting an SDK implementation right, so app developers can focus on creating engaging experiences.
+The AEP SDK offers an extension to quickly inspect, validate, debug data collection, and experiences for any mobile app using the AEP SDK. We built Assurance to do the heavy lifting of getting an SDK implementation right, so app developers can focus on creating engaging experiences.
 
 ## Quick Setup
 
@@ -21,14 +21,18 @@ pod 'AEPAssurance'
 Now that the backward compatible layer is installed, `ACPCore` can now be imported into the `AppDelegate` file allowing Assurance to be registered. Replace usage of `MobileCore.registerExtensions` with `ACPCore.registerExtensions` while including the Assurance extension.
 
 ```diff
++ import ACPCore
 + import AEPAssurance
 
 func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
 
-+  AEPAssurance.registerExtension()
-   MobileCore.registerExtensions([Lifecycle.self, Identity.self, Signal.self, ...], {
-      //...
-   })  
++  AEPAssurance.registerExtension()  
++  ACPCore.registerExtensions([Lifecycle.self, Identity.self, Signal.self, ...], {
++     //...
++  })
+-  MobileCore.registerExtensions([Lifecycle.self, Identity.self, Signal.self, ...], {
+-      //...
+-  })
 
   return true
 } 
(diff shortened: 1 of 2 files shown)


Documentation/Debugging.md
```diff
@@ -0,0 +1,46 @@
...
++ import ACPCore
++ import AEPAssurance
+
+func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
+
+-  MobileCore.registerExtensions([Lifecycle.self, Identity.self, Signal.self, ...], {
+-   //...
+-  })  
+
++  ACPCore.registerExtensions([Lifecycle.self, Identity.self, Signal.self, AEPAssurance.self, ...], {
```
shalehaha, line 33: does this really work? I don't think `AEPAssurance` is an extension class

nporter-adbe: mmmm. good catch, yeah I forgot it doesn't directly inherit from `ACPExtension`. let me fix this.

nporter-adbe: Will also fix in my 3rd party PR..

shalehaha: https://github.com/adobe/aepsdk-sample-app-ios/blob/main/Swift/AEPSampleApp/AppDelegate.swift#L54-L57
This is the code we have in the sample app

Documentation/Debugging.md
```diff
@@ -0,0 +1,42 @@
+# Implementation Validation with Assurance
+
+The AEP SDK offers an extension to inspect quickly, validate, debug data collection, and experiences for any mobile app using the AEP SDK. We built Assurance to do the heavy lifting of getting an SDK implementation right, so app developers can focus on creating engaging experiences.
```
sbenedicadb, line 3: `inspect quickly` -> `quickly inspect` (makes it lead into your list a little better)

Documentation/Debugging.md
```diff
@@ -0,0 +1,42 @@
...
+
+Now that the backward compatible layer is installed, `ACPCore` can now be imported into the `AppDelegate` file allowing Assurance to be registered. Replace usage of `MobileCore.registerExtensions` with `ACPCore.registerExtensions` while including the Assurance extension.
+
+```diff
++ import AEPAssurance
+
+func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
+
++  AEPAssurance.registerExtension()
+   MobileCore.registerExtensions([Lifecycle.self, Identity.self, Signal.self, ...], {
```
sbenedicadb, line 29: `ACPCore.registerExtensions`?

Documentation/GettingStarted.md
```diff
@@ -94,5 +94,5 @@ To download more examples of integrating the AEP Swift SDK, head over to the sam
 
 - Get familiar with the various APIs offered by the AEP SDK by checking out the [API usage documents](./Usage/). 
 - Integrate with the [ Experience Platform Extension](https://github.com/adobe/aepsdk-platform-ios). 
-- To leverage shared services offered by the AEP SDK, check out the [Services documentation](./Services/README.md).
+- Validate SDK implementation with [Assurance](./Debugging.md).
```
sbenedicadb, line 97: 👍

Documentation/Debugging.md
nporter-adbe: looks like we need to update the sample app.

sbenedicadb (approved):
looks good pending answer to my question

Documentation/Debugging.md
shalehaha: that was the only way I can get Assurance working. `start` method has to be called.

nporter-adbe: I think with this change: https://github.com/adobe/aepsdk-compatibility-ios/pull/5 it allows a more standard approach now.

nporter-adbe: I'll give this a test and update the sample app if needed.

shalehaha: I see, it does work. Didn't notice this change.
But still, it is something quite confusing.

nporter-adbe: I do agree it can be confusing, however, it's quite a big improvement from having to register the extension in the callback then start the SDK again.

Given the only change on this workflow is using ACPCore instead of MobileCore on registerExtensions, it would be optimal if we could somehow provide the backward compatible API under MobileCore. I attempted to create an Obj-C category on MobileCore in the compatibility layer, but this causes a circular dependency which caused compiling to fail when included from Cocoapods. Maybe we can take another swing at this.

PR #373, opened 2020-10-08
CodeCov Fixes

cdhoffmann:
Set target coverage to 85% and delta for reduced coverage to 5%

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -8,8 +8,12 @@ coverage:
   range: "70...90"
 
   status:
-    project: yes
-    patch: yes
+    project:
+      default:
+        informational: true
+    patch:
+      default:
+        informational: true
     changes: no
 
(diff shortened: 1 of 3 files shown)


codecov.yml
```diff
@@ -8,8 +8,12 @@ coverage:
   range: "70...90"
 
   status:
-    project: yes
-    patch: yes
+    project:
+      default:
+        informational: true
```
sbenedicadb, line 13: is the `informational` flag what prevents a decrease from causing a failure?

nporter-adbe: i hope so :D

cdhoffmann: That's what i'm wanting to test. It's supposed to stop any failure from happening. It will generate reports but not fail. But CI is down so i can't test.

PR #374, opened 2020-10-08
Increase timeout on testUnRegisterExtensionsSimple

nporter-adbe:


nporter-adbe:
@cdhoffmann has included this fix in his or #373

Issue #376, opened 2020-10-15
Add TrackAction/TrackState APIs


sbenedicadb:
- [x] Add public apis
- [x] Update compatibility layer

PR #377, opened 2020-10-15
#376 - adding tracking apis

sbenedicadb:
#376 - Adds tracking apis and tests

also updates all event names to be consistent with the c++ core


AEPCore/Sources/core/MobileCore+Tracking.swift
```diff
@@ -0,0 +1,51 @@
...
+
+/// Defines the public interface for the Tracking methods
+@objc
+public extension MobileCore {
+    /// Generates and dispatches a track action `Event`
+    /// - Parameters:
+    ///   - action: `String` representing the name of the action to be tracked
+    ///   - data: Dictionary of data to attach to the dispatched `Event`
+    @objc(trackAction:data:)
+    static func trackAction(_ action: String?, data: [String: Any]?) {
```
nporter-adbe, line 23: Any reason the first parameter names are omitted when calling the API? `_ action: String?` vs `action: String?`

nporter-adbe (approved):
nice! one question.

AEPCore/Sources/core/MobileCore+Tracking.swift
sbenedicadb: `action` is the last word in the method name, and also the parameter name.  i think in this case the label is implied, so that's why i omitted.  thoughts?

nporter-adbe: got it, makes sense to me 👍

shalehaha: static func track(action: String?, data: [String: Any]?) 
static func track(state: String?, data: [String: Any]?) 

How about this?

cdhoffmann:
I am actually observing some differences in naming right now while validating v5 vs Swift SDK in Griffon. Hold off on merging until i can verify that we are fixing these here.

PR #379, opened 2020-10-16
Use raw value instead of enum on syncIdentifiers

nporter-adbe:
@cdhoffmann pointed out we get an error encoding it with AnyCodable.
`AnyCodable - encode: Failed to encode Optional(AEPIdentity.MobileVisitorAuthenticationState)`


shalehaha:
@nporter-adbe 
any idea why we get this error since `MobileVisitorAuthenticationState` already conforms to `codable`?

nporter-adbe:
```
Reading config file at /Users/distiller/project/.swiftformat
/Users/distiller/project/AEPServices/Sources/ApplicationSystemInfoService.swift:106:1: warning: (indent) Indent code in accordance with the scope level.
SwiftFormat completed in 0.82s.
Source input did not pass lint check.
1/248 files require formatting.
```

I've tried:
`make format`

`make checkFormat`

`make lint-autocorrect`

shalehaha:
@nporter-adbe 
you need to pull from the latest dev branch. Also, the current CI is set up to always use the latest SwfitFormat tool, so you need to upgrade it locally.

nporter-adbe:
@shalehaha yes, AnyCodable can't handle any type that conforms to `Codable`, it attempts to encode/decode based on common `Codable` types. However, it's impossible for `AnyCodable` to know about every possible `Codable` type to attempt to decode.

For example, to decode a string it does
```swift
if let string = try? container.decode(String.self) {
   self.init(string)
}
```

However, if you imagine in the case of AutenticationStatus:
```swift
if let string = try? container.decode(MobileVisitorAuthenticationState.self) {
   self.init(string)
}
```

The above won't work as MobileVisitorAuthenticationState is defined in AEPIdentity, not in AEPCore.

One idea we could do to improve this is
1. For all Codable types in Core we add cases to handle them within AnyCodable
2. We allow extensions to supply a list of Codable types it would like AnyCodable to handle

cdhoffmann (approved):
thanks for fixing

PR #382, opened 2020-10-19
-fix wrong param name

sbenedicadb:


AEPCore/Sources/core/MobileCore.swift
```diff
@@ -95,8 +95,8 @@ public final class MobileCore: NSObject {
     /// Submits a generic event containing the provided IDFA with event type `generic.identity`.
     /// - Parameter identifier: the advertising identifier string.
     @objc(setAdvertisingIdentifier:)
-    public static func setAdvertisingIdentifier(adId: String?) {
```
nporter-adbe, line 98: I'd prefer to change the doc comment than the code, as it could break sample apps.

sbenedicadb: it does break sample code (and tests...), but i think `identifier` is a more correct name for the parameter than `adId`.  thoughts?

nporter-adbe: don't have much of a preference, I think they're about equal in my book. up to you.

sbenedicadb:
closing for deeper review of public apis

PR #388, opened 2020-10-21
Fix sync identifiers hit url

nporter-adbe:
Was adding Optional(9) instead of 9 to the sync identifiers hit url.


AEPIdentity/Sources/IdentityState.swift
```diff
@@ -346,7 +346,7 @@ class IdentityState {
         }
 
         if let ecid = identityResponse.ecid, !ecid.isEmpty {
-            let stringHint: String? = identityResponse.hint == nil ? nil : "\(String(describing: identityResponse.hint))"
+            let stringHint: String? = identityResponse.hint == nil ? nil : "\(String(describing: identityResponse.hint!))"
```
shalehaha, line 349: is it safe to force unwrap?

nporter-adbe: Yeah, in the expression we check if the hint is nil before force unwrapping.

shalehaha (approved):

PR #392, opened 2020-10-22
Flatten CustomIdentities in SharedState and Identity Refactor

nporter-adbe:
In Events and shared state coming from Identity, we were putting in our custom Swift type `CustomIdentity`, this format didn't match 100% with our v5 event format and also was causing issues when the data was being sent to Assurance as it was unable to deserialize the type.

Original event data:
```jsx
{
    blob = j8Odv6LonN4r3an7LhD3WZrU1bUpAkFkkiY1ncBR96t2PTI;
    lastsync = "1603311210.452279";
    locationhint = 9;
    mid = 90520654536133369506854802638635608086;
    visitoridslist =     (
        "AEPIdentity.CustomIdentity",
        "AEPIdentity.CustomIdentity"
    );
}
```

Updated event data, so now we flatten our `CustomIdentity` to a dictionary for easy reading in event data.
```jsx
{
    "visitoridslist": [
      {
        "id": "1234567",
        "authentication.state": 1,
        "id.origin": "d_cid_ic",
        "id.type": "idType1"
      }
    ],
    "mid": "47638672570484457245556507411373198888",
    "blob": "j8Odv6LonN4r3an7LhD3WZrU1bUpAkFkkiY1ncBR96t2PTI",
    "locationhint": "9"
  }
```

This PR also does some minor refactoring to reduce the number of times we need to decode a `DataEntity` when processing an `IdentityHit`.

### DIFF
--- a/AEPIdentity/Sources/MobileIdentities.swift
+++ b/AEPIdentity/Sources/MobileIdentities.swift
@@ -104,7 +104,7 @@ struct MobileIdentities: Codable {
         // So, there is no need to fetch the advertising identifier with advertisingidentifer namespace DSID_20914 separately.
         if let customVisitorIds = identitySharedState.value?[IdentityConstants.EventDataKeys.VISITOR_IDS_LIST] as? [[String: Any]] {
             // convert each `CustomIdentity` dictionary representations to a `UserID`, then remove any nil values
-            visitorIds.append(contentsOf: customVisitorIds.map { dictToUserID(dict: $0) }.compactMap { $0 })
+            visitorIds.append(contentsOf: customVisitorIds.map { MobileIdentities.dictToUserID(dict: $0) }.compactMap { $0 })
         }
 
         // push identifier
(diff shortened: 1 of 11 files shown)


AEPIdentity/Sources/MobileIdentities.swift
```diff
@@ -127,13 +127,13 @@ struct MobileIdentities: Codable {
...
     }
-}
 
-private extension CustomIdentity {
-    /// Converts a `CustomIdentity` into a `UserID`
+    /// Converts a `CustomIdentity` dictionary representation into a `UserID`
     /// - Returns: a `UserID` where the namespace is value, value is identifier, and type is integrationCode
-    func toUserID() -> UserID? {
-        guard let type = type, let identifier = identifier, !identifier.isEmpty else { return nil }
+    private func dictToUserID(dict: [String: Any]) -> UserID? {
```
shalehaha, line 133: can this be a static method?

shalehaha (approved):

AEPIdentity/Sources/MobileIdentities.swift
nporter-adbe: I suppose it could be, what's the use case?

shalehaha: just easier to understand the behavior of the method, it won't change any instance vars

cdhoffmann (approved):
looks good!

PR #421, opened 2020-10-23
Add XDM SharedState APIs

nporter-adbe:
This PR implements XDM shared state APIs as described in this doc: https://wiki.corp.adobe.com/pages/viewpage.action?spaceKey=adms&title=XDM+Shared+States

These APIs will be used for extensions to write XDM data to shared state, then the Edge extension will read this data and attach it to outgoing Platform events.

Note: The added tests are pretty light, I feel that we have adequate tests for SharedState logic, however, I've added a test to ensure that we are reading/writing to the correct SharedState instance based on the SharedStateType.

### DIFF
--- a/AEPCore/Sources/eventhub/ExtensionContainer.swift
+++ b/AEPCore/Sources/eventhub/ExtensionContainer.swift
@@ -107,6 +107,10 @@ extension ExtensionContainer: ExtensionRuntime {
     func createPendingSharedState(event: Event?) -> SharedStateResolver {
         return EventHub.shared.createPendingSharedState(extensionName: sharedStateName, event: event)
     }
+    
+    func createPendingXDMSharedState(event: Event?) -> SharedStateResolver {
+        return EventHub.shared.createPendingSharedState(extensionName: sharedStateName, event: event, sharedStateType: .xdm)
+    }
 
     func getSharedState(extensionName: String, event: Event?, barrier: Bool = true) -> SharedStateResult? {
         return EventHub.shared.getSharedState(extensionName: extensionName, event: event, barrier: barrier)
--- a/AEPCore/Sources/eventhub/ExtensionContainer.swift
+++ b/AEPCore/Sources/eventhub/ExtensionContainer.swift
@@ -25,6 +25,9 @@ class ExtensionContainer {
     /// The `SharedState` associated with the extension
     var sharedState: SharedState?
 
+    /// The XDM `SharedState` associated with the extension
+    var xdmSharedState: SharedState?
+
     var sharedStateName: String = "invalidSharedStateName"
 
     /// The extension's dispatch queue
@@ -55,10 +58,23 @@ class ExtensionContainer {
             }
 
             self.sharedState = SharedState(unwrappedExtension.name)
+            self.xdmSharedState = SharedState("xdm.\(unwrappedExtension.name)")
             self.sharedStateName = unwrappedExtension.name
             unwrappedExtension.onRegistered()
             self.eventOrderer.start()
             completion(nil)
+        }
+    }
+
+    /// Returns the corresponding `SharedState` for the given `SharedStateType`
+    /// - Parameter type: type of shared state to be retrieved
+    /// - Returns: The `SharedState` instance mapped to `type`
+    func sharedState(for type: SharedStateType) -> SharedState? {
+        switch type {
+        case .standard:
+            return sharedState
+        case .xdm:
+            return xdmSharedState
         }
     }
 }
@@ -94,6 +110,14 @@ extension ExtensionContainer: ExtensionRuntime {
 
     func getSharedState(extensionName: String, event: Event?, barrier: Bool = true) -> SharedStateResult? {
         return EventHub.shared.getSharedState(extensionName: extensionName, event: event, barrier: barrier)
+    }
+
+    func createXDMSharedState(data: [String: Any], event: Event?) {
+        EventHub.shared.createSharedState(extensionName: sharedStateName, data: data, event: event, sharedStateType: .xdm)
+    }
+
+    func getXDMSharedState(extensionName: String, event: Event?) -> SharedStateResult? {
+        return EventHub.shared.getSharedState(extensionName: extensionName, event: event, sharedStateType: .xdm)
     }
 
     func startEvents() {
--- a/AEPCore/Sources/eventhub/ExtensionContainer.swift
+++ b/AEPCore/Sources/eventhub/ExtensionContainer.swift
@@ -108,16 +108,16 @@ extension ExtensionContainer: ExtensionRuntime {
         return EventHub.shared.createPendingSharedState(extensionName: sharedStateName, event: event)
     }
 
-    func createPendingXDMSharedState(event: Event?) -> SharedStateResolver {
-        return EventHub.shared.createPendingSharedState(extensionName: sharedStateName, event: event, sharedStateType: .xdm)
-    }
-
     func getSharedState(extensionName: String, event: Event?, barrier: Bool = true) -> SharedStateResult? {
         return EventHub.shared.getSharedState(extensionName: extensionName, event: event, barrier: barrier)
     }
 
     func createXDMSharedState(data: [String: Any], event: Event?) {
         EventHub.shared.createSharedState(extensionName: sharedStateName, data: data, event: event, sharedStateType: .xdm)
+    }
+
+    func createPendingXDMSharedState(event: Event?) -> SharedStateResolver {
+        return EventHub.shared.createPendingSharedState(extensionName: sharedStateName, event: event, sharedStateType: .xdm)
     }
 
     func getXDMSharedState(extensionName: String, event: Event?) -> SharedStateResult? {
--- a/AEPCore/Sources/eventhub/ExtensionContainer.swift
+++ b/AEPCore/Sources/eventhub/ExtensionContainer.swift
@@ -107,7 +107,7 @@ extension ExtensionContainer: ExtensionRuntime {
     func createPendingSharedState(event: Event?) -> SharedStateResolver {
         return EventHub.shared.createPendingSharedState(extensionName: sharedStateName, event: event)
     }
-    
+
     func createPendingXDMSharedState(event: Event?) -> SharedStateResolver {
         return EventHub.shared.createPendingSharedState(extensionName: sharedStateName, event: event, sharedStateType: .xdm)
     }
--- a/AEPCore/Sources/eventhub/ExtensionRuntime.swift
+++ b/AEPCore/Sources/eventhub/ExtensionRuntime.swift
@@ -59,6 +59,14 @@ public protocol ExtensionRuntime {
     /// - Returns: a `SharedStateResolver` that should be called with the `SharedState` data when it is ready
     func createPendingSharedState(event: Event?) -> SharedStateResolver
 
+    /// Creates a pending XDM `SharedState` versioned at `event`
+    /// If `event` is nil, one of two behaviors will be observed:
+    /// 1. If this extension has not previously published a shared state, shared state will be versioned at 0
+    /// 2. If this extension has previously published a shared state, shared state will be versioned at the latest
+    /// - Parameter event: `Event` for which the `SharedState` should be versioned
+    /// - Returns: a `SharedStateResolver` that should be called with the `SharedState` data when it is ready
+    func createPendingXDMSharedState(event: Event?) -> SharedStateResolver
+
     /// Gets the `SharedState` data for a specified extension    
     /// - Parameters:
     ///   - extensionName: An extension name whose `SharedState` will be returned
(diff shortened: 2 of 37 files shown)


shalehaha:
Question: Why don't we just make xdm a new field in the current shared state?

nporter-adbe:
@shalehaha This has been discussed I believe, not sure if it's documented in the document. @emdobrin can you chime in on this?

shalehaha:
haveing two shared states will potentially impact rules engine and also girffon.

> I believe one reason is that we do not want pending shared states when using XDM shared states.
With the change we have made, now the last valid shared state is returned if the current status is pending. So this shouldn't be a problem.

emdobrin:
@shalehaha @nporter-adbe The main reason for having separate APIs is that we want to have a contract with the developers, the extension should only map XDM formatted data to these shared states and the associated Launch extension should declare the schema used on the client, so we can automatically map rules to those schemas. 

I think it would be nice to potentially display this data in a new tab/view in Griffon, so you can see all XDM shared states at once. Also, this will make it easier for the extensions to update only one at a time. There can be extensions having only XDM shared states or only regular shared states or both. 

If the shared state name causes an issue, we can think about a naming convention e.g.  com.adobe.mobile.identity.xdm

shalehaha:
@emdobrin The contract can be API or it can be the structure of the shared state, either way will work. 

>this will make it easier for the extensions to update only one at a time. 

This is the thing I think we need an agreement on first. For an extension that updates both normal shared state and xdm shared state, shouldn't we force them to update both at the same time, is there really a case where only one of them need to be updated?
My overall feeling is just that it is weird that each extension has two different shared state. It makes sense from the xdm's perspective, but from Core's side, I just want things to be as clean as possible.

emdobrin:
@shalehaha I think we will have a mix of extensions that handle both regular&xdm shared states (e.g. for maintaining the backwards compatibility) and only one type of shared states. For the first situation, it may be required to update both at the same time, but not necessarily.

We would prefer to keep them separated so that the Edge extension, Launch rules or third party extensions can fetch only the XDM data if needed. This way they don't need to take care of curating the additional keys, they can take the entire shared state or search for a known xdm path.

shalehaha:
@emdobrin 
I totally got what you mean. 
But one thing, it would be better to keep Core as simple as possible, not saying we couldn't have two different shared states, but it will become something super hard to maintain in the future. Especially you said they may not have the exact same data. 
Also from the extension's side, yes, they may only need to update one of the shared state. but the question is, should we give them such flexibility? I personally would like to force an extension to update both at the same time, it should be quite simple for an extension since it can always keep a copy of the current shared state.


I personally don't think there is anything that needs to change to the current shared state APIs, extensions can just add a `xmd` field to the shared state. And i don't think that will cause any problem to Edge extension, Launch rules or third party extensions, then just need to read the `xdm` filed from the shared state.

However, if you do think a API to get/set shared state can simplify things, my proposal is keeping only one shared state which contains both normal shared state and xdm shared state. So it can be structured like : 
```
{
"noraml": {....}
"xdm" : {....}
}
```
the updateSharedState API can accept an additional optional parameter for xdm shared state.
the getter method can be similar to this PR, or we can rename the original `getSharedState` to something more specific.

nporter-adbe:
I agree that we should keep Core clean. However, I do believe these APIs should exist in Core. If these were to live in the Edge extension, then any extension who wants to use XDM APIs would need to depend on the Edge extension, or at least reference it weakly not to introduce a hard dependency.

Regarding whether we have a separate SharedState instance or an additional `xdm` field in the existing SharedState, I don't think it makes much difference. The overhead for adding a new SharedState instance is pretty low, but at the same time, it doesn't really give us any additional flexibility that we would not get from just adding an `xdm` field. One concern about adding an `xdm` field is we are somewhat setting the precedent for putting a mix of data into one SharedState collection.

I'd prefer to have separate APIs for the XDM shared state. I prefer to keep our existing shared state APIs as simple as possible and not require all extensions to understand/know about XDM shared state, but rather keep XDM shared state its own thing and extensions that need to know about it can use the XDM APIs. One option is to have a new API such as `createSharedState(standardData: [String: Any], xdmData: [String: Any], event: Event?)` to enforce that extensions update both shared states at the same time.

shalehaha:
With one shared state, we can enforce everything being updated at the same time, so no multi-threading implications. That is one minor benefit of that.
The bottom line to me is that, unless there is a critical reason that we have to go with multiple shared states, we should just use one shared state, I believe it will save us from any potential issues can be caused by inconsistent between different shared states.
I'm okay with adding any necessary APIs to the Core.

nporter-adbe:
@shalehaha I can get on board with using just one shared state instance, let me update this PR.

nporter-adbe:
Closed in favor of: https://github.com/adobe/aepsdk-core-ios/pull/437

nporter-adbe:
Re-opening this PR based on our discussion with @pfransenadb.

@shalehaha @emdobrin please review.

AEPCore/Sources/eventhub/ExtensionRuntime.swift
```diff
@@ -66,4 +66,22 @@ public protocol ExtensionRuntime {
...
+    // MARK: - XDM Shared State
+
+    /// Creates a new XDM `SharedState` for this extension
+    /// If `event` is nil, one of two behaviors will be observed:
+    /// 1. If this extension has not previously published a shared state, shared state will be versioned at 0
+    /// 2. If this extension has previously published a shared state, shared state will be versioned at the latest
+    /// - Parameters:
+    ///   - data: Data for the `SharedState`
+    ///   - event: `Event` for which the `SharedState` should be versioned
+    func createXDMSharedState(data: [String: Any], event: Event?)
```
shalehaha, line 79: do we need `createPendingXDMSharedState` ?

nporter-adbe: I don't believe we have a use-case for pending XDM shared state. The Edge extension will just grab whatever is available for a given event.

@emdobrin can you confirm

pfransenadb: I think all of the requirements for shared state exist for the xdm counterpart.  Edge Extension is just one potential consumer of XDM Shared states, eventually we'd expect all shared states to be XDM -- as a concrete schema for all data will be super useful with rules and other cross-extension workflows.

emdobrin: If we have the same functionality as regular shared states where you can fetch latest when pending, I think we should add pending states for important data that always needs to be up-to-date. The extensions will decide if they need to use pending for sync use-cases.

nporter-adbe: Ok, the original spec didn't specify a pending XDM shared state API, but I'll update this PR to add one.

nporter-adbe: @shalehaha @emdobrin added pending APIs.

AEPCore/Sources/eventhub/ExtensionContainer.swift
```diff
@@ -92,10 +108,22 @@ extension ExtensionContainer: ExtensionRuntime {
         return EventHub.shared.createPendingSharedState(extensionName: sharedStateName, event: event)
     }
 
+    func createPendingXDMSharedState(event: Event?) -> SharedStateResolver {
```
emdobrin, line 111: Same here, let's group the xdm APIs together

AEPCore/Sources/eventhub/ExtensionRuntime.swift
```diff
@@ -59,11 +59,37 @@ public protocol ExtensionRuntime {
...
     /// - Parameters:
     ///   - extensionName: An extension name whose `SharedState` will be returned
     ///   - event: If not nil, will retrieve the `SharedState` that corresponds with the event's version, if nil will return the latest `SharedState`
     ///   - barrier: If true, the `EventHub` will only return `.set` if `extensionName` has moved past `event`
     /// - Returns: A `SharedStateResult?` for the requested `extensionName` and `event`
     func getSharedState(extensionName: String, event: Event?, barrier: Bool) -> SharedStateResult?
+
+    // MARK: - XDM Shared State
+
+    /// Creates a new XDM `SharedState` for this extension
```
emdobrin, line 80: Creates a new XDM `SharedState` for this extension. The data passed to this API needs to be mapped to known XDM mixins; if an extension uses multiple mixins, the current data for all of them should be provided when the XDM shared state is set.

AEPCore/Sources/eventhub/ExtensionRuntime.swift
```diff
@@ -59,11 +59,37 @@ public protocol ExtensionRuntime {
...
+    /// If `event` is nil, one of two behaviors will be observed:
+    /// 1. If this extension has not previously published a shared state, shared state will be versioned at 0
+    /// 2. If this extension has previously published a shared state, shared state will be versioned at the latest
+    /// - Parameters:
+    ///   - data: Data for the `SharedState`
+    ///   - event: `Event` for which the `SharedState` should be versioned
+    func createXDMSharedState(data: [String: Any], event: Event?)
+
+    /// Gets the XDM `SharedState` data for a specified extension
```
emdobrin, line 89: Gets the XDM `SharedState` data for a specified extension. If this extension populates multiple mixins in their shared state, all the data will be returned at once and it can be accessed using path discovery.

AEPCore/Sources/eventhub/SharedStateType.swift
```diff
@@ -0,0 +1,19 @@
...
+ OF ANY KIND, either express or implied. See the License for the specific language
+ governing permissions and limitations under the License.
+ */
+
+import Foundation
+
+/// Represents the `SharedState` types in `EventHub`
+enum SharedStateType {
+    case standard
+    case xdm
```
emdobrin, line 18: let's add some descriptions in here:
```
/// Represents the `SharedState` types in `EventHub` that can be used by an extension to share data with other extensions and for rules execution
enum SharedStateType {
 case standard // regular data, the key names and structure can be defined by each extension
 case xdm // data mapped on XDM mixins populated by an extension
}
```

AEPIdentity/Tests/TestHelpers/TestableExtensionRuntime.swift
```diff
@@ -50,14 +52,32 @@ class TestableExtensionRuntime: ExtensionRuntime {
         }
     }
 
+    func createPendingXDMSharedState(event: Event?) -> SharedStateResolver {
```
emdobrin, line 55: Same here, let's group the new API together

nporter-adbe:
@emdobrin thanks for the feedback, updated.

shalehaha:
@nporter-adbe 
remember to update all the related docs as well

nporter-adbe:
@shalehaha thanks, updated the building extensions doc. Any others I missed?

shalehaha:
https://github.com/adobe/aepsdk-core-ios/blob/main/Documentation/EventHub/DefinitionOfTerms.md, this page to explain standard vs xdm share state?
and also https://github.com/adobe/aepsdk-core-ios/blob/main/Documentation/EventHub/Contract.md#shared-state

Documentation/EventHub/BuildingExtensions.md
```diff
@@ -249,6 +249,56 @@ All extensions are provided a default API to read shared state from another exte
...
+/// If `event` is nil, one of two behaviors will be observed:
+/// 1. If this extension has not previously published a shared state, shared state will be versioned at 0
+/// 2. If this extension has previously published a shared state, shared state will be versioned at the latest
+/// - Parameters:
+///   - data: Data for the `SharedState`
+///   - event: `Event` for which the `SharedState` should be versioned
+func createXDMSharedState(data: [String: Any], event: Event?)
+```
+
+##### Creating and Updating an XDM Pending Shared State
```
emdobrin, line 272: rewording: a Pending XDM Shared State

Documentation/EventHub/BuildingExtensions.md
```diff
@@ -249,6 +249,56 @@ All extensions are provided a default API to read shared state from another exte
...
+
+In some cases, an extension may want to declare that its shared state is currently pending. For example, an extension may be doing some data manipulation, but in the meantime, the extension may invalidate its existing shared state and notify other extensions that the extension is currently working on providing a new shared state. This can be done with the API `func createPendingSharedState(event: Event?) -> SharedStateResolver`. This function creates a pending shared state versioned at an optional `Event` and returns a closure, which is to be invoked with your updated shared state data once available.
```
emdobrin, line 274: `func createPendingXDMSharedState ...` ... with your updated XDM shared state data once available.

Documentation/EventHub/DefinitionOfTerms.md
```diff
@@ -64,6 +64,13 @@ A **Shared State** that has been created, but not yet been populated by the **Ex
 A mechanism that allows **Extensions** to share state data with other Extensions. Any data existing in a shared state is valid  until it is either overwritten or removed by the owning extension. Shared states are owned by the **Event Hub**, but maintained by the Extension that owns them.
 
 --------------------------------------------------------------------------------
+
+### XDM Shared State
+
+A mechanism that allows **Extensions** to share state XDM compliant data with other Extensions. XDM shared states allow extensions allow the Edge extension to collect XDM data from various mobile extensions when needed and allow for the creation of XDM data elements to be used in Launch rules.
```
emdobrin, line 70: ...XDM shared states allow the Edge extension to collect XDM data ...

emdobrin (approved):
Few rewordings on the docs are needed, otherwise it looks good

PR #434, opened 2020-10-28
-fix date formatting and add tests

sbenedicadb:
#435


AEPServices/Sources/utility/Date+Format.swift
```diff
@@ -16,17 +16,24 @@ public extension Date {
...
+            return dateString
+        }
+        dateString.remove(at: index)
+        return dateString
     }
 
+    // e.g. - 2020-10-28T15:08:32-06:00
     func getISO8601Date() -> String {
-        let formatter = DateFormatter()
-        formatter.locale = Locale(identifier: "en_US_POSIX")
```
shalehaha, line 28: i'm a little bit concerned about removing the locale.

shalehaha: Is that handled by `ISO8601DateFormatter `?

AEPServices/Sources/utility/Date+Format.swift
```diff
@@ -16,17 +16,24 @@ public extension Date {
         return Int64(timeIntervalSince1970)
     }
 
+    // e.g. - 2020-10-28T15:08:32-0600
     func getRFC822Date() -> String {
-        let formatter = DateFormatter()
-        formatter.locale = Locale(identifier: "en_US_POSIX")
-        formatter.setLocalizedDateFormatFromTemplate("yyyy-MM-dd'T'HH:mm:ssZZZ")
-        return formatter.string(from: self)
+        let formatter = ISO8601DateFormatter()
```
yangyansong-adbe, line 21: Should we use `DateFormatter` instead of `ISO8601DateFormatter`?
`
When working with fixed format dates, such as RFC 3339, you set the dateFormat property to specify a format string. For most fixed formats, you should also set the locale property to a POSIX locale ("en_US_POSIX"), and set the timeZone property to UTC.
` 
-- from [Apple](https://developer.apple.com/documentation/foundation/dateformatter)

Here is my sample code: 
```Swift
let formatter = DateFormatter()
formatter.locale = Locale(identifier: "en_US_POSIX")
formatter.dateFormat = "yyyy-MM-dd'T'HH:mm:ssZZZ"
print(formatter.string(from: Date()))
//2020-10-29T12:52:42-0600
```

sbenedicadb:
@shalehaha quick! while it's green!

PR #436, opened 2020-11-02
syncronize Event processing

nporter-adbe:
Adds all event processing onto the eventHubQueue to fix any race conditions that could result in out of order Event dispatching.

### DIFF
--- a/AEPCore/Sources/eventhub/EventHub.swift
+++ b/AEPCore/Sources/eventhub/EventHub.swift
@@ -81,10 +81,12 @@ final class EventHub {
     /// Dispatches a new `Event` to the `EventHub`. This `Event` is sent to all listeners who have registered for the `EventType`and `EventSource`
     /// - Parameter event: An `Event` to be dispatched to listeners
     func dispatch(event: Event) {
-        // Set an event number for the event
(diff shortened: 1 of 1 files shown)


shalehaha (approved):

AEPCore/Sources/eventhub/EventHub.swift
```diff
@@ -195,25 +201,27 @@ final class EventHub {
     ///   - barrier: If true, the `EventHub` will only return `.set` if `extensionName` has moved past `event`
     /// - Returns: The `SharedState` data and status for the extension with `extensionName`
     func getSharedState(extensionName: String, event: Event?, barrier: Bool = true) -> SharedStateResult? {
-        guard let container = registeredExtensions.first(where: { $1.sharedStateName == extensionName })?.value, let sharedState = container.sharedState else {
-            Log.warning(label: LOG_TAG, "Unable to retrieve shared state for \(extensionName). No such extension is registered.")
-            return nil
-        }
+        return eventHubQueue.sync {
```
shalehaha, line 204: a little bit concerned about the performance implication but we can see if it will cause any problem

nporter-adbe:
any feedback @sbenedicadb?

PR #437, opened 2020-11-02
XDM shared state APIs

nporter-adbe:
A follow up to https://github.com/adobe/aepsdk-core-ios/pull/421

This PR implements XDM shared state APIs as described in this doc and based on feedback in PR #421: https://wiki.corp.adobe.com/pages/viewpage.action?spaceKey=adms&title=XDM+Shared+States

These APIs will be used for extensions to write XDM data to shared state, then the Edge extension will read this data and attach it to outgoing Platform events.

Note: The added tests are pretty light, I feel that we have adequate tests for SharedState logic, however, I've added a test to ensure that we are reading/writing to the correct SharedState instance based on the SharedStateType.

### DIFF
--- a/AEPCore/Mocks/TestableExtensionRuntime.swift
+++ b/AEPCore/Mocks/TestableExtensionRuntime.swift
@@ -56,12 +56,12 @@ public class TestableExtensionRuntime: ExtensionRuntime {
         }
         return mockedSharedStates["\(extensionName)"]
     }
-    
+
     public func createSharedState(data: [String : Any], xdmData: [String : Any], event: Event?) {
         createdSharedStates += [data]
         createdXdmSharedStates += [xdmData]
     }
-    
+
     public func getXDMSharedState(extensionName: String, event: Event?) -> SharedStateResult? {
         if let id = event?.id {
             return mockedXdmSharedStates["\(extensionName)-\(id)"] ?? mockedXdmSharedStates["\(extensionName)"]
--- a/AEPCore/Mocks/TestableExtensionRuntime.swift
+++ b/AEPCore/Mocks/TestableExtensionRuntime.swift
@@ -20,7 +20,9 @@ public class TestableExtensionRuntime: ExtensionRuntime {
     public var listeners: [String: EventListener] = [:]
     public var dispatchedEvents: [Event] = []
     public var createdSharedStates: [[String: Any]?] = []
+    public var createdXdmSharedStates: [[String: Any]?] = []
     public var mockedSharedStates: [String: SharedStateResult] = [:]
+    public var mockedXdmSharedStates: [String: SharedStateResult] = [:]
 
     public init() {}
 
@@ -53,6 +55,18 @@ public class TestableExtensionRuntime: ExtensionRuntime {
             return mockedSharedStates["\(extensionName)-\(id)"] ?? mockedSharedStates["\(extensionName)"]
         }
         return mockedSharedStates["\(extensionName)"]
+    }
+    
+    public func createSharedState(data: [String : Any], xdmData: [String : Any], event: Event?) {
+        createdSharedStates += [data]
+        createdXdmSharedStates += [xdmData]
+    }
+    
+    public func getXDMSharedState(extensionName: String, event: Event?) -> SharedStateResult? {
+        if let id = event?.id {
+            return mockedXdmSharedStates["\(extensionName)-\(id)"] ?? mockedXdmSharedStates["\(extensionName)"]
+        }
+        return mockedXdmSharedStates["\(extensionName)"]
     }
 
     public func startEvents() {}
--- a/AEPCore/Mocks/TestableExtensionRuntime.swift
+++ b/AEPCore/Mocks/TestableExtensionRuntime.swift
@@ -39,8 +39,9 @@ public class TestableExtensionRuntime: ExtensionRuntime {
         dispatchedEvents += [event]
     }
 
-    public func createSharedState(data: [String: Any], event _: Event?) {
+    public func createSharedState(data: [String : Any], xdmData: [String : Any]?, event: Event?) {
         createdSharedStates += [data]
+        createdXdmSharedStates += [xdmData]
     }
 
     public func createPendingSharedState(event _: Event?) -> SharedStateResolver {
@@ -55,11 +56,6 @@ public class TestableExtensionRuntime: ExtensionRuntime {
             return mockedSharedStates["\(extensionName)-\(id)"] ?? mockedSharedStates["\(extensionName)"]
         }
         return mockedSharedStates["\(extensionName)"]
-    }
-
-    public func createSharedState(data: [String : Any], xdmData: [String : Any], event: Event?) {
-        createdSharedStates += [data]
-        createdXdmSharedStates += [xdmData]
     }
 
     public func getXDMSharedState(extensionName: String, event: Event?) -> SharedStateResult? {
--- a/AEPCore/Sources/configuration/Configuration.swift
+++ b/AEPCore/Sources/configuration/Configuration.swift
@@ -56,7 +56,7 @@ class Configuration: Extension {
         if !config.isEmpty {
             let responseEvent = Event(name: CoreConstants.EventNames.CONFIGURATION_RESPONSE_EVENT, type: EventType.configuration, source: EventSource.responseContent, data: config)
             dispatch(event: responseEvent)
-            createSharedState(data: config, event: nil)
+            createSharedState(data: config, xdmData: nil, event: nil)
             // notify rules engine to load cached rules
             if let rulesURLString = config[ConfigurationConstants.Keys.RULES_URL] as? String {
                 rulesEngine.loadCachedRules(for: rulesURLString)
--- a/AEPCore/Sources/eventhub/EventHub.swift
+++ b/AEPCore/Sources/eventhub/EventHub.swift
@@ -161,8 +161,7 @@ final class EventHub {
             return
         }
 
-        let sharedStateData = [SharedStateType.standard.rawValue: data, SharedStateType.xdm.rawValue: xdmData]
-        sharedState.set(version: version, data: sharedStateData as [String: Any])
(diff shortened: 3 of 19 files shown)


AEPCore/Sources/eventhub/ExtensionContainer.swift
```diff
@@ -96,6 +96,14 @@ extension ExtensionContainer: ExtensionRuntime {
         return EventHub.shared.getSharedState(extensionName: extensionName, event: event, barrier: barrier)
     }
 
+    func createSharedState(data: [String: Any], xdmData: [String: Any], event: Event?) {
```
shalehaha, line 99: should we do it similar to https://github.com/adobe/aepsdk-core-ios/pull/437/files#diff-02bd1581c751b39213c003d305212478496f8a5290a1a877cef358a338c892f4R158, make xdmData optional and only keep one `createSharedState ` method ？

nporter-adbe: I opted to overload it for a few reasons
1. Provide an API to create shared state without needing to interact/understand what XDM data is
2. Reduce breaking changes/updates in other extensions by updating this API

thoughts?

shalehaha: either way, the extension developers need to know whether they need to deal with XDM data (by selecting which method to use vs using an optional parameter)
if we don't mind breaking the current extensions and do an update to all, actually I'd prefer to keep this new method to force them passing into both data even if one of them is empty.

nporter-adbe: updated.

shalehaha:
@pfransenadb do you want to quickly review the change in this PR?

pfransenadb (commented):
I don't have a problem combining XDM into the same shared state container.  My only concern is the requirement that both types of data (XDM and non) must be provided for each update (otherwise one would potentially end up blank).  It's probably an edge case, as most extensions should be supplying a version of each type of data...

AEPCore/Sources/eventhub/Extension.swift
```diff
@@ -81,9 +81,10 @@ public extension Extension {
     /// 2. If this extension has previously published a shared state, shared state will be versioned at the latest
     /// - Parameters:
     ///   - data: Data for the `SharedState`
+    ///   - xdmData: XDM data for the `SharedState`
     ///   - event: `Event` for which the `SharedState` should be versioned
-    func createSharedState(data: [String: Any], event: Event?) {
-        runtime.createSharedState(data: data, event: event)
+    func createSharedState(data: [String: Any], xdmData: [String: Any]?, event: Event?) {
```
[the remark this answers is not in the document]
nporter-adbe: This wasn't what was suggested by @shalehaha in the previous PR discussion, and the APIs in the PR is based on that discussion. @shalehaha thoughts?

AEPCore/Sources/eventhub/EventHub.swift
```diff
@@ -153,14 +153,15 @@ final class EventHub {
     /// - Parameters:
     ///   - extensionName: Extension whose `SharedState` is to be updated
     ///   - data: Data for the `SharedState`
+    ///   - xdmData: XDM data for the `SharedState`
     ///   - event: `Event` for which the `SharedState` should be versioned
-    func createSharedState(extensionName: String, data: [String: Any]?, event: Event?) {
+    func createSharedState(extensionName: String, data: [String: Any]?, xdmData: [String: Any]? = nil, event: Event?) {
```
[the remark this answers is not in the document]
nporter-adbe: Regarding breaking changes, I think we're OK with breaking changes since we're in beta.
@shalehaha would like to hear your thoughts too.

shalehaha:
@pfransenadb  @emdobrin 
The reason I don't want to make the parameter optional or provide another overload API, is 1) to reduce the chance a developer might make a mistake where they only update one of the shared states, and 2) also keep the Core API clean and simple.

AEPCore/Sources/eventhub/EventHub.swift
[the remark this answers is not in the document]
shalehaha: this -> we are maintaining the backwards compatibility only through the compatibility layer.

another question related to this, for any 3rd party extension, if they want to move to Platform, should we force them to depend on the new Swift API, or we allow them to keep building it with the compatibility layer. @emdobrin 

i'd like them to build the extension with the Swift Core directly (either in swift or obj-c). Compatibility layer is only for existing customer and 3rd party extensions.

AEPCore/Sources/configuration/Configuration.swift
```diff
@@ -56,7 +56,7 @@ class Configuration: Extension {
         if !config.isEmpty {
             let responseEvent = Event(name: CoreConstants.EventNames.CONFIGURATION_RESPONSE_EVENT, type: EventType.configuration, source: EventSource.responseContent, data: config)
             dispatch(event: responseEvent)
-            createSharedState(data: config, event: nil)
+            createSharedState(data: config, xdmData: nil, event: nil)
```
emdobrin, line 59: Let's revert this change.

AEPCore/Sources/eventhub/EventHub.swift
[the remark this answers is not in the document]
emdobrin: +1 to maintain the compatibility layer mainly for backwards compatibility purposes. New features should be added on top of swift core.

AEPCore/Sources/eventhub/Extension.swift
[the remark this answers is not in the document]
emdobrin: Let's keep the old API and add the other two for xdm support.

AEPCore/Sources/eventhub/ExtensionContainer.swift
```diff
@@ -96,6 +96,10 @@ extension ExtensionContainer: ExtensionRuntime {
         return EventHub.shared.getSharedState(extensionName: extensionName, event: event, barrier: barrier)
     }
 
+    func getXDMSharedState(extensionName: String, event: Event?) -> SharedStateResult? {
```
emdobrin, line 99: Should we provide a barrier for this API too?

AEPCore/Tests/EventHubTests/EventHubTests.swift
```diff
@@ -629,7 +633,7 @@ class EventHubTests: XCTestCase {
         eventHub.start()
 
         // test
-        eventHub.createSharedState(extensionName: "mockExtensionTwo", data: SharedStateTestHelper.ONE, event: nil)
+        eventHub.createSharedState(extensionName: "mockExtensionTwo", data: SharedStateTestHelper.ONE.standard, event: nil)
```
emdobrin, line 636: These tests shouldn't be changed if we overload the API. Let's add new tests for the new APIs.

AEPCore/Mocks/TestableExtensionRuntime.swift
```diff
@@ -37,8 +39,9 @@ public class TestableExtensionRuntime: ExtensionRuntime {
         dispatchedEvents += [event]
     }
 
-    public func createSharedState(data: [String: Any], event _: Event?) {
+    public func createSharedState(data: [String : Any], xdmData: [String : Any]?, event: Event?) {
```
emdobrin, line 42: if we are providing an API like this we should make both data and xdmData non-optionals.

AEPCore/Sources/eventhub/ExtensionRuntime.swift
```diff
@@ -48,8 +48,9 @@ public protocol ExtensionRuntime {
     /// 2. If this extension has previously published a shared state, shared state will be versioned at the latest
     /// - Parameters:
     ///   - data: Data for the `SharedState`
+    ///   - xdmData: XDM data for the `SharedState`
     ///   - event: `Event` for which the `SharedState` should be versioned
-    func createSharedState(data: [String: Any], event: Event?)
+    func createSharedState(data: [String: Any], xdmData: [String: Any]?, event: Event?)
```
emdobrin, line 53: Same comment here, if we want both data and xdmData to be provided they shouldn't be optional.

AEPCore/Tests/FunctionalTests/EventHubContractTests.swift
```diff
@@ -281,8 +281,8 @@ class EventHubContractTest: XCTestCase {
             startExpectation.fulfill()
         }
         wait(for: [startExpectation], timeout: 1)
-        ContractExtensionOne.runtime?.createSharedState(data: ["event":"first"], event: event1)
-        ContractExtensionOne.runtime?.createSharedState(data: ["event":"second"], event: event2)
+        ContractExtensionOne.runtime?.createSharedState(data: ["event":"first"], xdmData: nil, event: event1)
+        ContractExtensionOne.runtime?.createSharedState(data: ["event":"second"], xdmData: nil, event: event2)
```
emdobrin, line 285: Let's revert these changes and add new ones with xdm data

AEPLifecycle/Sources/Lifecycle.swift
```diff
@@ -39,7 +39,7 @@ public class Lifecycle: NSObject, Extension {
         registerListener(type: EventType.genericLifecycle, source: EventSource.requestContent, listener: receiveLifecycleRequest(event:))
 
         let sharedStateData = [LifecycleConstants.EventDataKeys.LIFECYCLE_CONTEXT_DATA: lifecycleState.computeBootData().toEventData()]
-        createSharedState(data: sharedStateData as [String: Any], event: nil)
+        createSharedState(data: sharedStateData as [String: Any], xdmData: nil, event: nil)
```
emdobrin, line 42: Let's revert these changes for now

nporter-adbe:
> Let's keep the old API and add the other two for xdm support.

@emdobrin most of your feedback is based on the premise that we will add separate APIs for the XDM shared state. Given the current state of our conversation, it doesn't sound like @shalehaha is onboard. I'm going to hold off on addressing your feedback until we can come to a consensus.

PR #446, opened 2020-11-10
Publish Identity XDM Shared State (ECID, CustomIdentities)

nporter-adbe:
This PR adds support for sharing the `identityMap` as part of Identity's shared state. Any time that we update the standard shared state we also update the XDM shared state with a XDM representation of the current Identity data.

Sample Identity XDM shared state:
```json
{
  "identityMap" : {
    "ECID" : [
      {
        "id" : "59725191566662510105001351686447723202",
        "authenticationState" : "ambiguous"
      }
    ],
    "test-type" : [
      {
        "id" : "test-identifier",
        "authenticationState" : "authenticated",
        "primary" : false
      }
    ]
  }
}
```

### DIFF
--- a/AEPCore/Sources/eventhub/Extension.swift
+++ b/AEPCore/Sources/eventhub/Extension.swift
@@ -115,6 +115,39 @@ public extension Extension {
         return runtime.getSharedState(extensionName: extensionName, event: event, barrier: true)
     }
 
+    // MARK: - XDM Shared State
+
+    /// Creates a new XDM SharedState for this extension.
+    /// The data passed to this API needs to be mapped to known XDM mixins; if an extension uses multiple mixins, the current data for all of them should be provided when the XDM shared state is set.
+    /// If `event` is nil, one of two behaviors will be observed:
+    /// 1. If this extension has not previously published a shared state, shared state will be versioned at 0
+    /// 2. If this extension has previously published a shared state, shared state will be versioned at the latest
+    /// - Parameters:
+    ///   - data: Data for the `SharedState`
+    ///   - event: `Event` for which the `SharedState` should be versioned
+    func createXDMSharedState(data: [String: Any], event: Event?) {
+        runtime.createXDMSharedState(data: data, event: event)
+    }
+
+    /// Creates a pending XDM `SharedState` versioned at `event`
+    /// If `event` is nil, one of two behaviors will be observed:
+    /// 1. If this extension has not previously published a shared state, shared state will be versioned at 0
+    /// 2. If this extension has previously published a shared state, shared state will be versioned at the latest
+    /// - Parameter event: `Event` for which the `SharedState` should be versioned
+    /// - Returns: a `SharedStateResolver` that should be called with the `SharedState` data when it is ready
+    func createPendingXDMSharedState(event: Event?) -> SharedStateResolver {
+        return runtime.createPendingXDMSharedState(event: event)
+    }
+
+    /// Gets the XDM SharedState data for a specified extension. If this extension populates multiple mixins in their shared state, all the data will be returned at once and it can be accessed using path discovery.
+    /// - Parameters:
+    ///   - extensionName: An extension name whose `SharedState` will be returned
+    ///   - event: If not nil, will retrieve the `SharedState` that corresponds with the event's version, if nil will return the latest `SharedState`
+    /// - Returns: A `SharedStateResult?` for the requested `extensionName` and `event`
+    func getXDMSharedState(extensionName: String, event: Event?) -> SharedStateResult? {
+        return runtime.getXDMSharedState(extensionName: extensionName, event: event)
+    }
+
     /// Called before each `Event` is processed by any `ExtensionListener` owned by this `Extension`
     /// Should be overridden by any extension that wants to control it's own event flow on a per event basis.
     /// - Parameter event: `Event` that will be processed next
--- a/AEPIdentity/Sources/IdentityMap.swift
+++ b/AEPIdentity/Sources/IdentityMap.swift
@@ -42,10 +42,10 @@ struct IdentityMap: Equatable {
     ///   - primary: Indicates if this identity is the preferred identity. It is used as a hint to help systems better organize how identities are queried.
     mutating func addItem(namespace: String,
                           id: String,
-                          authenticationState: MobileVisitorAuthenticationState? = nil,
+                          authenticationState: XDMAuthenticationState? = nil,
                           primary: Bool? = nil) {
         let item = IdentityItem(id: id,
-                                authenticationState: XDMAuthenticationState.authStateFromMobileAuthState(authState: authenticationState), primary: primary)
+                                authenticationState: authenticationState, primary: primary)
 
         if var namespaceItems = items[namespace] {
             if let index = namespaceItems.firstIndex(of: item) {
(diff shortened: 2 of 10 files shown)


AEPIdentity/Sources/IdentityMap.swift
```diff
@@ -0,0 +1,98 @@
...
+            return .loggedOut
+        default:
+            return .ambiguous
+        }
+    }
+}
+
+/// Defines a map containing a set of end user identities, keyed on either namespace integration code or the namespace ID of the identity.
+/// Within each namespace, the identity is unique. The values of the map are an array, meaning that more than one identity of each namespace may be carried.
+struct IdentityMap: Equatable {
```
nporter-adbe, line 33: Copied this over and made some minor changes such as renaming the auth state to `XDMAuthenticationState`.

AEPCore/Sources/eventhub/Extension.swift
```diff
@@ -115,6 +115,39 @@ public extension Extension {
         return runtime.getSharedState(extensionName: extensionName, event: event, barrier: true)
     }
 
+    // MARK: - XDM Shared State
```
nporter-adbe, line 118: Forgot to add these in the previous XDM shared state PR.

nporter-adbe:
@emdobrin @kevinlind need some feedback when time permits

AEPCore/Sources/eventhub/Extension.swift
emdobrin: Can you split this in a separate PR so we can update dev with this and merge the Identity changes in a feature branch?

AEPIdentity/Sources/IdentityMap.swift
```diff
@@ -0,0 +1,98 @@
...
+    /// Adds an `IdentityItem` to this map. If an item is added which shares the same `namespace` and `id` as an item
+    /// already in the map, then the new item replaces the existing item.
+    /// - Parameters:
+    ///   - namespace: The namespace for this identity
+    ///   - id: Identity of the consumer in the related namespace.
+    ///   - authenticationState: The state this identity is authenticated as for this observed ExperienceEvent.
+    ///   - primary: Indicates if this identity is the preferred identity. It is used as a hint to help systems better organize how identities are queried.
+    mutating func addItem(namespace: String,
+                          id: String,
+                          authenticationState: MobileVisitorAuthenticationState? = nil,
```
emdobrin, line 45: I suggest to keep these concepts separated, and use XDMAuthenticationState here, this way we can use this API for pure xdm objects and maybe have some helper functions for conversions.

AEPIdentity/Sources/IdentityMap.swift
```diff
@@ -0,0 +1,98 @@
...
+                          primary: Bool? = nil) {
+        let item = IdentityItem(id: id,
```
emdobrin, line 47: We should check that namespace and id are non-empty strings.

AEPIdentity/Sources/IdentityProperties.swift
```diff
@@ -63,6 +63,24 @@ struct IdentityProperties: Codable {
...
+        var identityMap = IdentityMap()
+
+        // Add ECID
+        if let ecidString = ecid?.ecidString {
+            identityMap.addItem(namespace: IdentityConstants.XDM.Keys.ECID, id: ecidString)
+        }
+
+        for customerId in customerIds ?? [] {
+            guard let type = customerId.type, let id = customerId.identifier else { continue }
+            identityMap.addItem(namespace: type, id: id, authenticationState: customerId.authenticationState, primary: false)
```
emdobrin, line 78: This looks good for now for customer visitor ids. We are still confirming with the services team that this is supported for all customer id name, but we can revise this later when we have more details.

We will need to handle IDFA differently, which is currently stored in the customerIds list.

AEPIdentity/Sources/IdentityState.swift
```diff
@@ -145,14 +145,17 @@ class IdentityState {
     ///   - response: the response data if any
     ///   - eventDispatcher: a function which when invoked dispatches an `Event` to the `EventHub`
     ///   - createSharedState: a function which when invoked creates a shared state for the Identity extension
-    func handleHitResponse(hit: IdentityHit, response: Data?, eventDispatcher: (Event) -> Void, createSharedState: ([String: Any], Event?) -> Void) {
+    ///   - createXDMSharedState: a function which when invoked creates a XDM shared state for the Identity extension
+    func handleHitResponse(hit: IdentityHit, response: Data?, eventDispatcher: (Event) -> Void, createSharedState: ([String: Any], Event?) -> Void,
+                           createXDMSharedState: ([String: Any], Event?) -> Void) {
```
emdobrin, line 150: This would not be required if this path is never hit after the migration.

nporter-adbe: Good catch, makes sense given my other PR now.

AEPCore/Sources/eventhub/Extension.swift
nporter-adbe: Created https://github.com/adobe/aepsdk-core-ios/pull/450 and changed the target for this PR to the xdm feature branch. I'll get the XDM feature branch sync'd with dev once the change to dev land.

AEPIdentity/Tests/IdentityMapTests.swift
```diff
@@ -0,0 +1,274 @@
...
+        XCTAssertEqual(1, spaceItems?.count)
+        XCTAssertEqual("id", spaceItems?[0].id)
+        XCTAssertEqual("ambiguous", spaceItems?[0].authenticationState?.rawValue)
+        XCTAssertFalse(spaceItems?[0].primary ?? true)
+
+        let unknown = identityMap.getItemsFor(namespace: "unknown")
+        XCTAssertNil(unknown)
+    }
+
+    func testAddItems() {
```
kevinlind, line 38: Add a few tests for empty namespace and id to help keep the test coverage up. 😄

nporter-adbe: Added :)

nporter-adbe:
@emdobrin @kevinlind feedback addressed, let me know if we're good to merge :)

kevinlind (approved):
Looks good for now. As @emdobrin pointed out, the IdentityMap namespace for IDFA is wrong, but we can update that later when we know the correct value.

AEPIdentity/Tests/IdentityStateTests.swift
```diff
@@ -868,6 +879,8 @@ class IdentityStateTests: XCTestCase {
         // test
         state.processPrivacyChange(event: event, createSharedState: { _, _ in
             sharedStateExpectation.fulfill()
+        }, createXDMSharedState: { _, _ in
+            xdmSharedStateExpectation.fulfill()
```
emdobrin, line 883: We should also have some e2e tests to verify the values stored in the xdm shared state are correct after syncIdentifiers call or setPrivacyStatus changes, I am ok if you want to add that in a separate pr.

emdobrin (commented):
@nporter-adbe changes look great, just added a suggestion for testing.

AEPIdentity/Tests/IdentityStateTests.swift
nporter-adbe: Already merged :)
I'll get these additional tests in when we determine the final format of the namespace.

PR #447, opened 2020-11-12
Disable Identity Hit Queuing if Edge is present

nporter-adbe:
This PR adds an API to the EventHub to check if an extension is registered. Then it uses this API to determine if we should disable hit queuing in Identity.

### DIFF
--- a/AEPCore/Sources/eventhub/EventHub.swift
+++ b/AEPCore/Sources/eventhub/EventHub.swift
@@ -268,6 +268,13 @@ final class EventHub {
         Log.debug(label: LOG_TAG, "Shared state created for \(EventHubConstants.NAME) with version \(version) and data: \n\(data as AnyObject)")
     }
 
+    /// Determines if there exists a registered extension with `extensionName`
+    /// - Parameter extensionName: The name of the extension to be looked up
+    /// - Returns: Returns true if this extension is currently registered with the `EventHub`, otherwise false
+    func isExtensionRegistered(extensionName: String) -> Bool {
+        return registeredExtensions.first(where: {$0.value.sharedStateName == extensionName}) != nil
+    }
+
     // MARK: - Private
 
     /// Gets the appropriate `SharedState` for the provided `extensionName` and `event`
--- a/AEPCore/Sources/eventhub/EventHub.swift
+++ b/AEPCore/Sources/eventhub/EventHub.swift
@@ -268,13 +268,6 @@ final class EventHub {
         Log.debug(label: LOG_TAG, "Shared state created for \(EventHubConstants.NAME) with version \(version) and data: \n\(data as AnyObject)")
     }
 
-    /// Determines if there exists a registered extension with `extensionName`
-    /// - Parameter extensionName: The name of the extension to be looked up
-    /// - Returns: Returns true if this extension is currently registered with the `EventHub`, otherwise false
-    func isExtensionRegistered(extensionName: String) -> Bool {
-        return registeredExtensions.first(where: {$0.value.sharedStateName == extensionName}) != nil
-    }
-
     // MARK: - Private
 
     /// Gets the appropriate `SharedState` for the provided `extensionName` and `event`
--- a/AEPCore/Tests/EventHubTests/EventHubTests.swift
+++ b/AEPCore/Tests/EventHubTests/EventHubTests.swift
@@ -1094,45 +1094,4 @@ class EventHubTests: XCTestCase {
         // verify
         validateSharedState(EventHubTests.MOCK_EXTENSION_NAME, event, "one", .xdm)
     }
-
-    // MARK: isExtensionRegistered() Tests
-
-    /// Ensures that false is returned when an non-registered extension is passed in
-    func testIsExtensionRegisteredInvalid() {
-        // setup
-        eventHub.start()
-
-        // test & verify
-        XCTAssertFalse(eventHub.isExtensionRegistered(extensionName: "com.adobe.test.extension"))
-    }
-
-    /// Ensures that true is returned when a registered extension name is passed in
-    func testIsExtensionRegisteredValid() {
-        // setup
-        eventHub.start()
-
-        // test & verify
-        // MockExtension is registred in setup()
-        XCTAssertTrue(eventHub.isExtensionRegistered(extensionName: MockExtension(runtime: TestableExtensionRuntime()).name))
-    }
-
-    /// Tests that isExtensionsRegistered returns the correct value for an extension that is registered, then unregistered
-    func testIsExtensionRegisteredReturnsTrueThenFalseAfterUnregistration() {
-        // setup
-        eventHub.start()
-        let expectation = XCTestExpectation(description: "MockExtension is unregistered")
-
-        // test & verify
-        // MockExtension is registred in setup()
-        XCTAssertTrue(eventHub.isExtensionRegistered(extensionName: MockExtension(runtime: TestableExtensionRuntime()).name))
-
-        // unregistered MockExtension
-        eventHub.unregisterExtension(MockExtension.self) { (error) in
-            XCTAssertNil(error)
-            XCTAssertFalse(self.eventHub.isExtensionRegistered(extensionName: MockExtension(runtime: TestableExtensionRuntime()).name))
-            expectation.fulfill()
-        }
-
-        wait(for: [expectation], timeout: 0.5)
-    }
 }
--- a/AEPCore/Tests/EventHubTests/EventHubTests.swift
+++ b/AEPCore/Tests/EventHubTests/EventHubTests.swift
@@ -1117,7 +1117,7 @@ class EventHubTests: XCTestCase {
     }
 
     /// Tests that isExtensionsRegistered returns the correct value for an extension that is registered, then unregistered
-    func testIsExtensionRegisteredRoundTrip() {
+    func testIsExtensionRegisteredReturnsTrueThenFalseAfterUnregistration() {
         // setup
         eventHub.start()
         let expectation = XCTestExpectation(description: "MockExtension is unregistered")
--- a/AEPCore/Tests/EventHubTests/EventHubTests.swift
+++ b/AEPCore/Tests/EventHubTests/EventHubTests.swift
@@ -1094,4 +1094,45 @@ class EventHubTests: XCTestCase {
         // verify
         validateSharedState(EventHubTests.MOCK_EXTENSION_NAME, event, "one", .xdm)
     }
+
+    // MARK: isExtensionRegistered() Tests
+
+    /// Ensures that false is returned when an non-registered extension is passed in
+    func testIsExtensionRegisteredInvalid() {
+        // setup
+        eventHub.start()
+
+        // test & verify
+        XCTAssertFalse(eventHub.isExtensionRegistered(extensionName: "com.adobe.test.extension"))
+    }
+
+    /// Ensures that true is returned when a registered extension name is passed in
+    func testIsExtensionRegisteredValid() {
+        // setup
+        eventHub.start()
+
+        // test & verify
+        // MockExtension is registred in setup()
+        XCTAssertTrue(eventHub.isExtensionRegistered(extensionName: MockExtension(runtime: TestableExtensionRuntime()).name))
+    }
+
+    /// Tests that isExtensionsRegistered returns the correct value for an extension that is registered, then unregistered
+    func testIsExtensionRegisteredRoundTrip() {
+        // setup
+        eventHub.start()
+        let expectation = XCTestExpectation(description: "MockExtension is unregistered")
+
+        // test & verify
+        // MockExtension is registred in setup()
+        XCTAssertTrue(eventHub.isExtensionRegistered(extensionName: MockExtension(runtime: TestableExtensionRuntime()).name))
+
+        // unregistered MockExtension
+        eventHub.unregisterExtension(MockExtension.self) { (error) in
+            XCTAssertNil(error)
+            XCTAssertFalse(self.eventHub.isExtensionRegistered(extensionName: MockExtension(runtime: TestableExtensionRuntime()).name))
+            expectation.fulfill()
+        }
+
+        wait(for: [expectation], timeout: 0.5)
+    }
 }
--- a/AEPIdentity/Sources/Identity.swift
+++ b/AEPIdentity/Sources/Identity.swift
@@ -33,11 +33,11 @@ import Foundation
             return
         }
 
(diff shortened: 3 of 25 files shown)


nporter-adbe:
I have an idea of how to make this a bit cooler via the HitProcessor, going to close and re-open tomorrow.

AEPIdentity/Sources/Identity.swift
```diff
@@ -224,4 +224,9 @@ import Foundation
 
         return privacyStatus == .optedOut
     }
+
+    /// - Returns: Returns true if the Edge extension is registered, false otherwise
+    private func isEdgeRegistered() -> Bool {
+        return isExtensionRegistered(extensionName: IdentityConstants.SharedStateKeys.EDGE)
```
nporter-adbe, line 230: A little concerned about calling this for each hit. Maybe we can use a "local static var" to reduce the number of times this invokes EventHub APIs.

```swift
    private func isEdgeRegistered() -> Bool {
       struct Holder {
           static var edgeRegistered
        }

       if Holder.edgeRegistered { return true }
       Holder.edgeRegistered = isExtensionRegistered(extensionName: IdentityConstants.SharedStateKeys.EDGE)
       return Holder.edgeRegistered
}
```

thoughts?

nporter-adbe:
@emdobrin @kevinlind @shalehaha need some feedback when time permits

AEPCore/Sources/eventhub/EventHub.swift
```diff
@@ -268,6 +268,13 @@ final class EventHub {
         Log.debug(label: LOG_TAG, "Shared state created for \(EventHubConstants.NAME) with version \(version) and data: \n\(data as AnyObject)")
     }
 
+    /// Determines if there exists a registered extension with `extensionName`
+    /// - Parameter extensionName: The name of the extension to be looked up
+    /// - Returns: Returns true if this extension is currently registered with the `EventHub`, otherwise false
+    func isExtensionRegistered(extensionName: String) -> Bool {
+        return registeredExtensions.first(where: {$0.value.sharedStateName == extensionName}) != nil
+    }
```
emdobrin, line 276: I would prefer not to expose this API yet as we don't encourage inter-extension dependencies. Another way to check that the edge extension is registered is by checking core's shared state.

AEPIdentity/Sources/Identity.swift
emdobrin: Ideally all the extensions should be registered before start so if we do that for the first event the Edge extension should be registered by that time. 
In addition to that, for subsequent launches we can check if the IdentityMap exists which would indicate the migration happened already once and we don't need to check the edge extension again.
We will not support downgrades to Identity direct once the migration happened.

AEPIdentity/Sources/Identity.swift
```diff
@@ -33,7 +33,7 @@ import Foundation
             return
         }
 
-        let hitQueue = PersistentHitQueue(dataQueue: dataQueue, processor: IdentityHitProcessor(responseHandler: handleNetworkResponse(hit:responseData:)))
+        let hitQueue = PersistentHitQueue(dataQueue: dataQueue, processor: IdentityHitProcessor(responseHandler: handleNetworkResponse(hit:responseData:), isEdgeRegistered: isEdgeRegistered))
```
emdobrin, line 36: I would prefer not to change the HitQueue for this use-case and handle everything in the Extension, this can allow us to remove this code in the future since it will not be used for Edge use-cases.

emdobrin (commented):
@nporter-adbe can you please create a feature branch in upstream repo for these changes as we are still exploring this concept?

AEPCore/Sources/eventhub/EventHub.swift
nporter-adbe: We already have use-cases where this API can be useful for example in Identity checking if Audience is present, and I believe there will be some in Analytics as well. I'd prefer not to have each extension duplicate logic for checking if an extension is registered (which will end up happening if we don't have an API provided) as parsing through the dictionary can be error-prone IMO.

AEPIdentity/Sources/Identity.swift
nporter-adbe: The addition of `func shouldQueue(entity: DataEntity) -> Bool` to the hit queue has applications outside of the Edge use case, for example, an extension can decide not to queue hits if the latest configuration contains an opted-out privacy status. IMO, this is a clean and protocol-driven approach to determining if a hit should be queued or not rather than loading up conditional statements to determine if a hit should be queued.

If someone using the hit queue doesn't really _want_ the new functionality they can simplify just implement `shouldQueue` to return true each time. This kind of pattern is seen throughout iOS, for example, https://developer.apple.com/documentation/uikit/uiviewcontroller/1621502-shouldperformseguewithidentifier

```swift
func shouldQueue(entity: DataEntity) -> Bool {
   return true
}
```

AEPIdentity/Sources/Identity.swift
nporter-adbe: I'm not quite sure I understand the feedback here. This PR specifically adds functionality to disable Identity from sending hits directly to the Identity service if Edge is present. I'd prefer not to have more logic in here than needed with checking if Edge is present, then keeping track of if a migration has occurred. 

It appears to me checking if Edge is present is sufficient enough to determine if Identity should queue/not queue hits. Thoughts?

nporter-adbe:
@emdobrin created a branch named `xdm` and changed the base of this PR, also responded to your comments.

AEPIdentity/Sources/Identity.swift
emdobrin: This approach does not cover the use-case where the migration happened, but in the next app launch the Edge extension is not registered anymore. In this case we should not fallback to the Identity direct path.

nporter-adbe: In the case that the migration has happened but Edge is not registered, what would be the behavior of the Identity extension? Do we fallback to not doing anything? When would this scenario occur?

AEPIdentity/Sources/IdentityHitProcessor.swift
```diff
@@ -44,6 +47,11 @@ class IdentityHitProcessor: HitProcessing {
         }
     }
 
+    func shouldQueue(entity: DataEntity) -> Bool {
+        // only queue hits if Edge is not registered
+        return !isEdgeRegistered()
```
kevinlind, line 52: We should log something here to indicate the hit was dropped.

AEPCore/Tests/EventHubTests/EventHubTests.swift
```diff
@@ -1094,4 +1094,45 @@ class EventHubTests: XCTestCase {
...
+        // setup
+        eventHub.start()
+
+        // test & verify
+        // MockExtension is registred in setup()
+        XCTAssertTrue(eventHub.isExtensionRegistered(extensionName: MockExtension(runtime: TestableExtensionRuntime()).name))
+    }
+
+    /// Tests that isExtensionsRegistered returns the correct value for an extension that is registered, then unregistered
+    func testIsExtensionRegisteredRoundTrip() {
```
kevinlind, line 1120: nit: "RoundTrip" isn't very self explanatory. Maybe `testIsExtensionRegisteredReturnsTrueThenFalseAfterUnregistration` ? I know, verbose, but it's more descriptive.

AEPIdentity/Sources/Identity.swift
nporter-adbe: Another thought I had is that even if we save a flag or check for the identityMap, it doesn't cover the case where the user uninstalls the app.
1. App launch with Identity + Edge
2. Uninstall the app
3. App launch with only Identity, now we can't determine if the migration happened so we'll fall back on the legacy Identity service, which could lead to inconsistent SDK behavior based on if the user has installed/re-installed. Should this new install be using the Identity service while other users could be going through Edge?

I think that changing the behavior of an extension based on the prior existence of an extension could be error-prone and lead to strange/inconsistent behavior. Unless there is a common workflow where the customer is going to use Edge then remove Edge, I'm not sure we should handle this case.

AEPIdentity/Sources/Identity.swift
shalehaha: should we even not initialize hitQueue when we know edge is enabled?
The current codes work, but it'd be better if can refactor the code to a less intrusive way where we don't need to check `isEdgeRegistered` at several different places.

nporter-adbe: We could probably de-init the hitQueue once we know Edge is present. One of the reasons I went with injecting the check into the hit queue was to make it easily testable. If we keep all the logic for de-init'ing the hitQueue in Identity.swift it is only really testable via functional tests (maybe that's ok). But I preferred to have a solution that would lend itself better to unit testing.
thoughts?

AEPIdentity/Sources/Identity.swift
nporter-adbe: @emdobrin any more thoughts on this?

nporter-adbe:
@shalehaha @emdobrin any more thoughts on this? should we do this in a simpler more direct way of just setting the hit queue to nil or is the current approach acceptable?

AEPIdentity/Sources/Identity.swift
emdobrin: The case where the Edge extension is not registered can occur on app update, maybe the Launch Edge extension is uninstalled or the registration code is removed for any reason. In this case we should update the identifiers in the IdentityMap, but they will not be synced with any service until the Edge extension is registered again.
For the app uninstall scenario you described above, I don't think we can catch that, but we should start thinking about a better representation of this dependency in Launch UI.

AEPCore/Sources/eventhub/EventHub.swift
emdobrin: If we don't expose this API at this point, this allows us to re-evaluate the Edge dependency check later on for Identity and other extensions if needed. Analytics Edge doesn't use this approach as it is a new extension.

AEPIdentity/Sources/Identity.swift
nporter-adbe: @emdobrin I'm not sure it makes more sense to not initialize the DB if not used. How would we not initialize the DB in this case? The DB is initialized in the init for Identity, at that point we don't know definitively whether or not the Edge extension is present or not.

If we never initialize the DB it could miss the case where the customer migrates to Edge but has existing Identity hits in the queue that need to go out. The proposed solution allows for existing hits that have been queued before Edge to be sent out, while not allowing additional new hits to be queued if Edge is present.

AEPCore/Sources/eventhub/EventHub.swift
nporter-adbe: I chatted with @shalehaha over the break and he's ok with exposing this API as well. It's not clear to me why we would need to re-evaluate the dependency check later on.

nporter-adbe:
@emdobrin updated the PR to remove the new event hub API and use the event hub shared state instead.

AEPIntegrationTests/IdentityIntegrationTests.swift
```diff
@@ -70,6 +73,29 @@ class IdentityIntegrationTests: XCTestCase {
         wait(for: [requestExpectation], timeout: 1)
     }
 
+    /// No network request should be made since the Edge extension is registered
+    func testSyncIdentifiersEdgeIsRegistered() {
```
emdobrin, line 77: Can we move this in a separate file, e.g. EdgeIdentityIntegrationTests? I expect we will add more integration tests for Edge use-cases later on.

AEPIntegrationTests/IdentityIntegrationTests.swift
```diff
@@ -37,13 +37,16 @@ class IdentityIntegrationTests: XCTestCase {
...
     }
 
-    func initExtensionsAndWait() {
+    func initExtensionsAndWait(registerMockEdge: Bool = false) {
         let initExpectation = XCTestExpectation(description: "init extensions")
         MobileCore.setLogLevel(.trace)
-        MobileCore.registerExtensions([Identity.self, Lifecycle.self, Signal.self]) {
+        var extensions = [Identity.self, Lifecycle.self, Signal.self] as! [Extension.Type]
+        if registerMockEdge {
+            extensions = [Identity.self, Lifecycle.self, Signal.self, MockEdgeExtension.self] as! [Extension.Type]
```
emdobrin, line 47: extensions.append(MockEdgeExtension.self)

AEPIdentity/Sources/Identity.swift
emdobrin: I agree with Jiabin here that we should only check the presence of Edge ext once when the processing starts instead of per event and potentially de-init the queue in that case since it will not be used. 
With regards to the previously queued hits, since this is an async process and Identity can process those hits after some new events were sent to Edge, we can run into inconsistent situations where the identifiers are overwritten, so I would prefer to sync whatever we have latest to Edge directly. I will look for a confirmation on this scenario though.

AEPIdentity/Sources/IdentityHitProcessor.swift
```diff
@@ -44,6 +47,15 @@ class IdentityHitProcessor: HitProcessing {
         }
     }
 
+    func shouldQueue(entity: DataEntity) -> Bool {
+        // only queue hits if Edge is not registered
+        let shouldQueue = isEdgeRegistered()
+        if !shouldQueue {
+            Log.debug(label: "\(LOG_TAG):shouldQueue", "Dropping hit with id: '\(entity.uniqueIdentifier)' because Edge is present.")
```
emdobrin, line 54: This log is a bit misleading as the hit is just not queued in Identity anymore, but the data is still going to be processed. We should log a message at the beginning of the processing saying that `AEP Edge is enabled, sending identifiers in XDM format` or something similar.

AEPIntegrationTests/IdentityIntegrationTests.swift
emdobrin: Similar comment here, we should have more e2e tests check the shared states are always up-to-date when public APIs are called, and that no network calls are going out.

nporter-adbe: Moved to another file, regarding checking shared state... our integration tests currently don't check shared state so this is something we'd need to look into for future integration tests.

AEPIdentity/Sources/Identity.swift
nporter-adbe: I'm a little concerned to set the queue to nil given an async task could be running. Not sure if that would lead to any memory issues. I've added an optimization to stop calling `getSharedState` once we know Edge is present. We should get the best of both worlds.

PR #453, opened 2020-11-18
Peek and Remove n for DataQueue

nporter-adbe:
Adds the ability to peek and remove n for the DataQueue.

### DIFF
--- a/AEPServices/Sources/dataqueue/SQLiteDataQueue.swift
+++ b/AEPServices/Sources/dataqueue/SQLiteDataQueue.swift
@@ -17,7 +17,6 @@ import Foundation
 ///    - `DataEntity` objects inside this queue will be persisted in SQLite database automatically
 ///    - the database operations is performed in series
 class SQLiteDataQueue: DataQueue {
-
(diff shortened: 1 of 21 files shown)


AEPServices/Sources/dataqueue/SQLiteDataQueue.swift
```diff
@@ -116,7 +109,7 @@ class SQLiteDataQueue: DataQueue {
                 disconnect(database: connection)
             }
             let deleteRowStatement = """
-            DELETE FROM \(SQLiteDataQueue.TABLE_NAME) order by id limit 1;
+            DELETE FROM \(SQLiteDataQueue.TABLE_NAME) order by id limit \(n);
```
shalehaha, line 112: nit: https://github.com/adobe/aepsdk-core-ios/pull/453/files#diff-4d8f660abc14b917eaee3faf76596aada2a9b34f04f6fa34a9b49be88ccc978cR79, the sql here are all uppercased, so it'd be better to keep them consistent.

shalehaha (approved):

PR #461, opened 2020-11-24
fix queue problem

shalehaha:
## Types of changes

## Checklist:

### DIFF
--- a/AEPServices/Sources/utility/hitprocessor/PersistentHitQueue.swift
+++ b/AEPServices/Sources/utility/hitprocessor/PersistentHitQueue.swift
@@ -63,7 +63,9 @@ public class PersistentHitQueue: HitQueuing {
             guard !self.suspended else { return }
             guard let hit = self.dataQueue.peek() else { return } // nothing left in the queue, stop processing
 
-            self.processor.processHit(entity: hit, completion: { [weak self] success in
+            
+            let semaphore = DispatchSemaphore(value: 0)
+            self.processor.processHit(entity: hit, completion: { [weak self] success in                
                 if success {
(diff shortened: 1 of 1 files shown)


shalehaha:
@nporter-adbe @cdhoffmann @sbenedicadb 
Let me know if there is a better way to fix this.

AEPServices/Sources/utility/hitprocessor/PersistentHitQueue.swift
```diff
@@ -63,7 +63,9 @@ public class PersistentHitQueue: HitQueuing {
             guard !self.suspended else { return }
             guard let hit = self.dataQueue.peek() else { return } // nothing left in the queue, stop processing
 
-            self.processor.processHit(entity: hit, completion: { [weak self] success in
+            
+            let semaphore = DispatchSemaphore(value: 0)
```
shalehaha, line 67: need to block the current dispatchqueue until getting the respsonse

nporter-adbe: Yeah, I can see how the bug could occur... I think this is probably the best solution, potentially using an `OperationQueue` from the Swift library could help us here but I don't think that's required.

nporter-adbe (approved):

nporter-adbe:
The only way I can see this bug happening is if `beginProcessing` is called while we are already processing hits. Another fix might to be only call `processNextHit` if we're not already suspended.

shalehaha:
@nporter-adbe I thought about using `suspended`, but it could mess things up with privacy status change. Not sure if introducing another bool flag similar to `suspended` can solve this problem.

Issue #459, opened 2020-11-23
couldn't change app id if the first one provided is invalid


shalehaha:
When I install the app with an invalid app id, SDK will keep retrying the download. Even after I update the app id to a correct one and relaunch the app, it will still retry download with the old invalid one(we cached the app id), so everything is hosed because of that.

PR #462, opened 2020-11-24
Fix persisted app id on launch

nporter-adbe:
Addresses: https://github.com/adobe/aepsdk-core-ios/issues/459

### DIFF
--- a/AEPCore/Sources/configuration/Configuration.swift
+++ b/AEPCore/Sources/configuration/Configuration.swift
@@ -46,11 +46,6 @@ class Configuration: Extension {
         registerListener(type: EventType.configuration, source: EventSource.requestContent, listener: receiveConfigurationRequest(event:))
         registerListener(type: EventType.lifecycle, source: EventSource.responseContent, listener: receiveLifecycleResponse(event:))
 
-        // If we have an appId stored in persistence, kick off the configureWithAppId event
-        if let appId = appIdManager.loadAppId(), !appId.isEmpty {
-            dispatchConfigurationRequest(data: [ConfigurationConstants.Keys.JSON_APP_ID: appId])
-        }
-
         configState.loadInitialConfig()
         let config = configState.environmentAwareConfiguration
         if !config.isEmpty {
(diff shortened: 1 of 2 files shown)


AEPCore/Sources/configuration/Configuration.swift
```diff
@@ -46,11 +46,6 @@ class Configuration: Extension {
         registerListener(type: EventType.configuration, source: EventSource.requestContent, listener: receiveConfigurationRequest(event:))
         registerListener(type: EventType.lifecycle, source: EventSource.responseContent, listener: receiveLifecycleResponse(event:))
 
-        // If we have an appId stored in persistence, kick off the configureWithAppId event
-        if let appId = appIdManager.loadAppId(), !appId.isEmpty {
```
shalehaha, line 50: `loadAppId ` also loads the id from the manifest, so we couldn't simply remove it. May need to directly use `loadAppIdFromPersistence() or  loadAppIdFromManifest()` here.

nporter-adbe: good point, fixed.

AEPCore/Tests/FunctionalTests/Configuration/ConfigurationLifecycleResponseTests.swift
```diff
@@ -80,36 +80,4 @@ class ConfigurationLifecycleResponseTests: XCTestCase {
         mockSystemInfoService.property = nil
     }
 
-    /// Tests that app id is loaded from manifest on normal startup after being persisted
-    func testHandleValidAppidFromManifestPersisted() {
```
shalehaha, line 84: should we keep this test?

nporter-adbe: I don't believe so. The first part of the test ensures that the app is loaded from the manifest, this part is still valid. But the second half of the tests removes the app if from the manifest and checks if the app is is loaded from persistence which is no longer valid. 

Maybe we can keep the first half of the test? I'd assume we already have a test for that case.

shalehaha (approved):

PR #464, opened 2020-11-30
get rid of semaphore in the PersistentHitQueue, use async way to handle respsonse

shalehaha:


shalehaha:
closing this PR, will learn and see if we can leverage dispatch group

PR #465, opened 2020-12-02
Don't encode push id

nporter-adbe:
We no longer want to encode the push ID, this PR removes encoding for push IDs.
https://jira.corp.adobe.com/browse/AMSDK-10812


AEPIdentity/Sources/PushIDManager.swift
```diff
@@ -75,16 +75,15 @@ struct PushIDManager: PushIDManageable {
     /// Compares the provided newPushId against the one in data store (if exists)
     /// - Parameter newPushId: the new push identifier as a string, not hashed yet
```
kevinlind, line 76: Update comment, remove "not hashed yet"

kevinlind (commented):
Can you update the comment in IdentityProperties.swift for var `pushIdentifier`, line 28?
Also, is the `SHA256.swift` utility class used anywhere else? Can we remove it?

nporter-adbe:
@kevinlind the SHA256 file is still used by the Identity extension, but it probably doesn't make much sense to have it named SHA256. I've replaced that file with an extension on `Data`.

nporter-adbe:
@emdobrin @kevinlind need a look here.

kevinlind (approved):

AEPIdentity/Sources/PushIDManager.swift
```diff
@@ -48,7 +48,7 @@ struct PushIDManager: PushIDManageable {
     }
 
     mutating func updatePushId(pushId: String?) {
```
emdobrin, line 50: I think we can use `pushId: String` here as we are doing the check in IdentityState syncIdentifiers

AEPCore/Sources/core/MobileCore.swift
```diff
@@ -125,8 +125,7 @@ public final class MobileCore: NSObject {
     /// - Parameter deviceToken: the device token for push notifications
     @objc(setPushIdentifier:)
     public static func setPushIdentifier(_ deviceToken: Data?) {
-        let hexString = SHA256.hexStringFromData(input: deviceToken as NSData?)
-        let data = [CoreConstants.Keys.PUSH_IDENTIFIER: hexString]
+        let data = [CoreConstants.Keys.PUSH_IDENTIFIER: deviceToken?.hexDescription ?? ""]
```
emdobrin, line 128: We should add some e2e tests that cover this change, if not already present.

AEPCore/Tests/Data+HexStringTests.swift
```diff
@@ -0,0 +1,36 @@
...
+ governing permissions and limitations under the License.
+ */
+
+@testable import AEPCore
+import XCTest
+
+class Data_HexStringTests: XCTestCase {
+
+    /// Tests that the string is properly encoded
+    func testHexStringFromDataHappy() {
```
emdobrin, line 18: let's add a test which covers all hex characters `0123456789ABCDEF`

emdobrin: Also, we should consider adding migration tests for push tokens to make sure they don't change (ACP->AEP).

AEPCore/Sources/core/MobileCore.swift
nporter-adbe: We have tests here:
https://github.com/adobe/aepsdk-core-ios/blob/70d941c71e9d1b5fac6b525618cfdc4af4514d03/AEPCore/Tests/MobileCore%2BIdentityTests.swift#L78
and here:
https://github.com/adobe/aepsdk-core-ios/blob/70d941c71e9d1b5fac6b525618cfdc4af4514d03/AEPIntegrationTests/IdentityIntegrationTests.swift#L153

AEPIdentity/Sources/PushIDManager.swift
nporter-adbe: We need to be able to pass `nil` in the case we are removing the push id in case of setting to opt out.

AEPCore/Tests/Data+HexStringTests.swift
nporter-adbe: We are not currently migrating the push ID as we don't migrate it from v4 ->v5. Since the customer usually sets push id on every launch I don't think it needs to be migrated.

I've added a longer test which should cover all the hex characters.

PR #472, opened 2020-12-17
Adding UI services

ravjain-adb:
## Description

- Adding UIServices with FullscreenMessage.
- Added `createFullscreenMessage` function to create the message with protocol `FullScreenMessageUiInterface`
- Added `MessageMonitor` object in UIService to handle whether any message is displayed or not.
- `FullScreenMessage` uses `FullscreenUIHandler` to handle `show`, `dismiss` and `openUrl`
- `FullscreenListenerInterface` is used to listen for callbacks from `FullscreenMessage`.

## Types of changes

## Checklist:

### DIFF
--- a/AEPServices/Sources/ui/MessageMonitor.swift
+++ b/AEPServices/Sources/ui/MessageMonitor.swift
@@ -1,0 +1,55 @@
+/*
+ Copyright 2020 Adobe. All rights reserved.
+ This file is licensed to you under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License. You may obtain a copy
+ of the License at http://www.apache.org/licenses/LICENSE-2.0
+ 
+ Unless required by applicable law or agreed to in writing, software distributed under
+ the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ OF ANY KIND, either express or implied. See the License for the specific language
+ governing permissions and limitations under the License.
+ */
+
+
+/*
+ This class is used to monitor if an UI message is displayed at some point in time,
+ currently this applies for full screen and alert messages.
+ The status will be exposed to the core through the UIService.
+ @see UIHandling.isMessageDisplayed()
+ */
+
+import Foundation
+
+public class MessageMonitor {
+
+    private var isMessageDisplayed = false
+    private let messageQueue = DispatchQueue(label: "com.adobe.uiservice.messagemonitor")
+
+    /*
+     Sets the isMessageDisplayed flag on true so other UI messages will not be displayed
+     in the same time.
+     */
+    func displayed() {
+        messageQueue.async {
+            self.isMessageDisplayed = true
+        }
+    }
+
+    /*
+     Sets the isMessageDisplayed flag on false enabling other messages to be displayed
+     */
+    func dismissed() {
+        messageQueue.async {
+            self.isMessageDisplayed = false
+        }
+    }
+
+    /*
+     Returns current status of the isMessageDisplayed flag
+     */
+    func isDisplayed() -> Bool {
+        return messageQueue.sync {
+            isMessageDisplayed
+        }
+    }
+}
--- a/AEPServices/Sources/ui/MessageMonitor.swift
+++ b/AEPServices/Sources/ui/MessageMonitor.swift
@@ -10,43 +10,34 @@
  governing permissions and limitations under the License.
  */
 
-
+import Foundation
 
 ///This class is used to monitor if an UI message is displayed at some point in time,
 ///currently this applies for full screen and alert messages.
- The status will be exposed to the core through the UIService.
- @see UIHandling.isMessageDisplayed()
- */
-
-import Foundation
-
+///The status will be exposed to the core through the UIService.
+///@see UIService.isMessageDisplayed()
 public class MessageMonitor {
 
     private var isMessageDisplayed = false
     private let messageQueue = DispatchQueue(label: "com.adobe.uiservice.messagemonitor")
 
-    /*
-     Sets the isMessageDisplayed flag on true so other UI messages will not be displayed
-     in the same time.
-     */
+    /// Sets the isMessageDisplayed flag on true so other UI messages will not be displayed
+    /// in the same time.
     func displayed() {
         messageQueue.async {
             self.isMessageDisplayed = true
         }
     }
 
-    /*
-     Sets the isMessageDisplayed flag on false enabling other messages to be displayed
-     */
+    
+    /// Sets the isMessageDisplayed flag on false enabling other messages to be displayed
     func dismissed() {
         messageQueue.async {
             self.isMessageDisplayed = false
         }
     }
 
-    /*
-     Returns current status of the isMessageDisplayed flag
-     */
+    /// Returns current status of the isMessageDisplayed flag
     func isDisplayed() -> Bool {
         return messageQueue.sync {
             isMessageDisplayed
--- a/AEPServices/Sources/ui/MessageMonitor.swift
+++ b/AEPServices/Sources/ui/MessageMonitor.swift
@@ -3,7 +3,7 @@
  This file is licensed to you under the Apache License, Version 2.0 (the "License");
  you may not use this file except in compliance with the License. You may obtain a copy
  of the License at http://www.apache.org/licenses/LICENSE-2.0
- 
+
  Unless required by applicable law or agreed to in writing, software distributed under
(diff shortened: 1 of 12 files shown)


AEPServices/Sources/ui/fullscreen/FullScreenMessageUiInterface.swift
```diff
@@ -0,0 +1,27 @@
...
+ Unless required by applicable law or agreed to in writing, software distributed under
+ the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ OF ANY KIND, either express or implied. See the License for the specific language
+ governing permissions and limitations under the License.
+ */
+
+import Foundation
+
+/// UI service interface defining a fullscreen message
+public protocol FullScreenMessageUiInterface {
```
sbenedicadb, line 16: seems like we are going back and forth between `Fullscreen` and `FullScreen`. let's pick one as as a group and be consistent with it.

AEPServices/Sources/ui/MessageMonitor.swift
```diff
@@ -0,0 +1,45 @@
...
+ governing permissions and limitations under the License.
+ */
+
+import Foundation
+
+///This class is used to monitor if an UI message is displayed at some point in time,
+///currently this applies for full screen and alert messages.
+///The status will be exposed to the core through the UIService.
+///@see UIService.isMessageDisplayed()
+public class MessageMonitor {
```
sbenedicadb, line 19: all of the classes and protocols are public. we should be considering what access level each one needs

AEPServices/Sources/ui/MessageMonitor.swift
```diff
@@ -0,0 +1,45 @@
...
+
+    private var isMessageDisplayed = false
+    private let messageQueue = DispatchQueue(label: "com.adobe.uiservice.messagemonitor")
+
+    /// Sets the isMessageDisplayed flag on true so other UI messages will not be displayed
+    /// in the same time.
+    func displayed() {
```
sbenedicadb, line 26: probably should have this method be named `display` and the next one be `dismiss` for clarity

AEPServices/Sources/ui/UIServiceInterface.swift
```diff
@@ -0,0 +1,43 @@
...
+
+/// Interface for displaying alerts, local notifications, and fullscreen web views
+public protocol UIServiceInterface {
+
+    /// Creates a fullscreen message.
+    /// WARNING: This API consumes HTML/CSS/JS using an embedded browser control.
+    /// This means it is subject to all the risks of rendering untrusted web pages and running untrusted JS.
+    /// Treat all calls to this API with caution and make sure input is vetted for safety somewhere.
+    ///
+    /// @param html               String html content to be displayed with the message
```
sbenedicadb, line 23: let's make sure we convert the comments to swift-style documentation

AEPServices/Sources/ui/UIService.swift
```diff
@@ -0,0 +1,35 @@
...
+ Unless required by applicable law or agreed to in writing, software distributed under
+ the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ OF ANY KIND, either express or implied. See the License for the specific language
+ governing permissions and limitations under the License.
+ */
+
+import Foundation
+import UIKit
+
+public class UIService: NSObject, UIServiceInterface {
```
sbenedicadb, line 16: need docs for this class

AEPServices/Sources/ui/fullscreen/FullScreenUIHandler.swift
```diff
@@ -0,0 +1,205 @@
...
+
+    /// Show the message, and call the OnShow() on the listener. Will only show the message if no other
+    /// fullscreen message is showing as of then.
+    func show() {
+        if monitor.isDisplayed() {
+            Log.debug(label: self.LOG_PREFIX, "Full screen message couldn't be displayed, another message is displayed at this time")
+            return
+        }
+
+        DispatchQueue.main.async {
```
sbenedicadb, line 49: we should add some code to the `FullscreenListenerInterface` that allows the developer to suppress showing a message.  seems like it should be added here?

shalehaha:
@ravjain-adb I hope we can simplify the code here. For v5, because we have cpp and objc layers, we have to define tons of interfaces and services for the sake of that, and it made our code actually quite complicated. But for swift, we only have one layer, so I think there are several places we can simplify things. Here are some of my ideas:
1. We can get rid of UIService, extensions can directly create fullscreen message using the FullScreenMessage class.
2. Change MessageMonitor to a service, it is the one that owns a state. And for now, it only tracks if there is any messge being display, but later I think it is a good place where we can allow app injecting custom behaviors for messages.
3. Merge FullScreenMessage and FullScreenUIHandler class.
4. Remove FullscreenListenerInterface, we can add some new onShow/onDismiss methods to the  FullScreenMessage class which accepts closures as parameters.

AEPServices/Sources/ui/MessageMonitor.swift
sbenedicadb: we also need to consider which ones need to be exposed to obj-c w/ `@objc`

sbenedicadb (changes requested):
looking good.  need to have a little more discussion on some of it.

cdhoffmann:
> @ravjain-adb I hope we can simplify the code here. For v5, because we have cpp and objc layers, we have to define tons of interfaces and services for the sake of that, and it made our code actually quite complicated. But for swift, we only have one layer, so I think there are several places we can simplify things. Here are some of my ideas:
> 
> 1. We can get rid of UIService, extensions can directly create fullscreen message using the FullScreenMessage class.
> 2. Change MessageMonitor to a service, it is the one that owns a state. And for now, it only tracks if there is any messge being display, but later I think it is a good place where we can allow app injecting custom behaviors for messages.
> 3. Merge FullScreenMessage and FullScreenUIHandler class.
> 4. Remove FullscreenListenerInterface, we can add some new onShow/onDismiss methods to the  FullScreenMessage class which accepts closures as parameters.

+1

AEPServices/Sources/ui/UIServiceInterface.swift
```diff
@@ -0,0 +1,43 @@
...
+ Unless required by applicable law or agreed to in writing, software distributed under
+ the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ OF ANY KIND, either express or implied. See the License for the specific language
+ governing permissions and limitations under the License.
+ */
+
+import Foundation
+
+/// Interface for displaying alerts, local notifications, and fullscreen web views
+public protocol UIServiceInterface {
```
cdhoffmann, line 16: Although protocols are technically interfaces, it's best practice in Swift to not name protocols interface. I would just remove the `interface` from all the protocols you've created. I know it makes naming the implementations tricker but just look at how we name our other services.

cdhoffmann (changes requested):

AEPServices/Sources/ui/MessageMonitor.swift
```diff
@@ -0,0 +1,45 @@
...
+
+    /// Sets the isMessageDisplayed flag on false enabling other messages to be displayed
+    func dismissed() {
+        messageQueue.async {
+            self.isMessageDisplayed = false
+        }
+    }
+
+    /// Returns current status of the isMessageDisplayed flag
+    func isDisplayed() -> Bool {
```
cdhoffmann, line 40: Since this doesn't have any parameters or side effects, a computed property would probably be cleaner.

cdhoffmann (changes requested):

AEPServices/Sources/ui/fullscreen/FullScreenUIHandler.swift
```diff
@@ -0,0 +1,205 @@
...
+                try FileManager.default.removeItem(atPath: tempHTMLFilePath)
+            } catch {
+                Log.debug(label: self.LOG_PREFIX, "Unable to dismiss \(error)")
+            }
+        }
+    }
+
+    /// Opens the resource at the URL specified.
+    func openUrl(url: String) {
+        if !url.isEmpty {
```
cdhoffmann, line 142: No need for the empty check if you will be guarding on the URL initializer because the initializer returns nil if the string is empty. So the guard suffices. Also, we do not use `URL.init(...)`. Simply use the `URL(string: ...)`

`guard let url = URL(string: url) else { return }`

sbenedicadb:
merging this into a feature branch so we can play around with it while we're working out the kinks

PR #478, opened 2020-12-22
Change swift version to 5.1 

cdhoffmann:


shalehaha:
@cdhoffmann  how about this one https://github.com/adobe/aepsdk-core-ios/blob/main/.swift-version?

shalehaha (approved):

PR #486, opened 2021-01-05
Allow per-hit computation of retry interval

nporter-adbe:
The Edge extension needs to have different retry intervals based on the network response. We'll need to make changes to the persistent hit queue to support variable retry intervals for different hits.

The Edge extension will use the new delegate function to determine the retry interval per hit.


AEPIdentity/Sources/IdentityHitProcessor.swift
```diff
@@ -15,7 +15,6 @@ import Foundation
 class IdentityHitProcessor: HitProcessing {
     private let LOG_TAG = "IdentityHitProcessor"
 
-    let retryInterval = TimeInterval(30)
```
emdobrin, line 18: It would be nice to keep both, so an extension can provide a retryInterval per entity if needed and also override the default retryInterval for all entities. The retry per entity should have priority if both of them are provided.

nporter-adbe: I'm not a fan of providing both. I find it a bit confusing to implement both IMO. If an implementer would like a default they can handle that within the delegate function. In essence, I'd prefer to have a single source of truth for the retry interval rather than having two, then determine priority.

shalehaha (approved):

Issue #432, opened 2020-10-28
shared state .barrier flag causes event not being processed in time


shalehaha:

nporter-adbe:
@shalehaha spent some time trying to reproduce this and couldn't get this behavior to occur. I'm not sure if we should close this issue or have someone who has experienced this issue to give it a try at reproducing the issue.

shalehaha:
Let's assume:
Event1: Configuration Update App id event
Event10: Lifecycle Start event
Event11: Eventhub Shared State update event

Configuration extension processes event1 and it's shared state is pending 
Lifecycle extension receives event10 but is blocked since configuration shared state is pending 
Configuration extension finish processing of event1, update its shared state
EventHub dispatch event11
Lifecycle extensions resume on event11, but when it gets configuration shared state with .barrier flag, the result is still pending. This is because the Configuration extension has not passed through event 10 yet.

... so now, unless another event comes in, lifecycle extension will just be hanging there.

shalehaha:
https://github.com/adobe/aepsdk-core-ios/pull/487 is just a temporary solution, will need more discussions later. So keep this task open.

PR #487, opened 2021-01-06
change the default value of .barrier to false

shalehaha:
A temporary solution for https://github.com/adobe/aepsdk-core-ios/issues/432


yangyansong-adbe:
@nporter-adbe  @jiabingeng   ignore the SonarCloud reports for now, I'm still testing it.

nporter-adbe (approved):

PR #490, opened 2021-01-07
Make HitQueuing+PrivacyStatus public

emdobrin:
## Description

Make HitQueuing+PrivacyStatus extension public to be used by other extensions outside of Core. To be used by Edge.

## Types of changes

## Checklist:

### DIFF
--- a/AEPCore/Sources/configuration/HitQueuing+PrivacyStatus.swift
+++ b/AEPCore/Sources/configuration/HitQueuing+PrivacyStatus.swift
@@ -13,10 +13,10 @@
 import AEPServices
 import Foundation
 
-extension HitQueuing {
+public extension HitQueuing {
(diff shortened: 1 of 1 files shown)


nporter-adbe:
This extension is currently under the Identity target; thus, it will only be available if you import Identity. I think it would be more ideal to move this to Core, so an extension that would like to use this utility extension doesn't need to explicitly depend/import Identity but rather core.

I considered that it might be a good fit under services, but we need the privacy status which is defined in core.

emdobrin:
@nporter-adbe it is currently under configuration so I believe Core should bring that in already.

nporter-adbe:
@emdobrin ah my mistake... Looks like the actual class is in core but then the unit tests are still under identity. It was originally under Identity then moved to core. Must have missed moving the unit tests too.

jiabingeng (approved):

shalehaha (approved):

PR #493, opened 2021-01-07
Improvements to UI service with full screen message

ravjain-adb:
## Description

- Fullscreen message ui handler is combined together.
- Adding new Global UI Messaging for end user to listen for current message lifecycle.
- Merging the UIService and messageMonitor together.
- Changes based on comments in #472 

## Types of changes

## Checklist:


AEPServices/Sources/ServiceProvider.swift
```diff
@@ -114,15 +113,10 @@ public class ServiceProvider {
         }
     }
 
-    public var uiService: UIServiceInterface {
+    public var messageMonitor: MessageMonitor {
```
nporter-adbe, line 116: nit: Can we keep the pattern of `service`  being the suffix for the variable name and the type?

AEPServices/Sources/ui/GlobalUIMessaging.swift
```diff
@@ -0,0 +1,31 @@
...
+ Unless required by applicable law or agreed to in writing, software distributed under
+ the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ OF ANY KIND, either express or implied. See the License for the specific language
+ governing permissions and limitations under the License.
+ */
+
+import Foundation
+
+/// Global UI Message delegate which is used to listen for current message lifecycle events
+@objc(AEPGlobalUIMessaging) public protocol GlobalUIMessaging {
```
nporter-adbe, line 16: In the doc comments, I see the mention of `message`, but don't see them in the function definition.

I'd be in favor of having all 3 protocol functions pass in a `message`, otherwise, it could be difficult to gain context into which messages are showing, dismissing, etc.

AEPServices/Sources/ui/GlobalUIMessaging.swift
```diff
@@ -0,0 +1,31 @@
...
+    func onShow()
+
+    /// Invoked when the any message is dismissed
+    /// - Parameters:
+    ///     - message: UIMessaging message that is being dismissed
+    func onDismiss()
+
+    /// Used to find whether messages should be shown or not
+    /// - Returns: true if message needs to be shown else false
+    func showMessage() -> Bool
```
nporter-adbe, line 30: Can we rename this to `shouldShowMessage`? I find that to be a bit clearer as its asking rather than doing.

AEPServices/Sources/ui/UIApplication+Window.swift
```diff
@@ -0,0 +1,16 @@
...
+//  AEPServices
+//
+//  Created by ravjain on 1/11/21.
+//  Copyright © 2021 Adobe. All rights reserved.
+//
+
+import Foundation
+import UIKit
+
+public extension UIApplication {
```
nporter-adbe, line 12: We've discussed this a bit, and we'd prefer to not expose public extensions on types unless we really need to as to not interfere with customers' potential existing extensions.

Is there any way we can make this internal?

AEPServices/Sources/ui/UIApplication+Window.swift
```diff
@@ -0,0 +1,16 @@
+//
+//  File.swift
```
nporter-adbe, line 2: Copyright header needs updating.

AEPServices/Sources/ui/alert/AlertMessaging.swift
```diff
@@ -12,16 +12,17 @@
 
 import Foundation
 
-/// UI service interface defining a fullscreen message
-public protocol FullScreenMessageUiInterface {
+/// Fullscreen message lifecycle event listener
+@objc(AEPAlertMessaging) public protocol AlertMessaging {
+    /// Invoked on positive button clicks
+    func OnPositiveResponse()
```
nporter-adbe, line 18: The first letter of the function name should be lowercase.
`onPositiveResponse`

AEPServices/Sources/ui/alert/AlertMessaging.swift
```diff
@@ -12,16 +12,17 @@
...
 
-    /// Display the fullscreen message
-    func show()
+    /// Invoked on negative button clicks
+    func OnNegativeResponse()
```
nporter-adbe, line 21: Lowercase first letter

AEPServices/Sources/ui/fullscreen/FullscreenMessage.swift
```diff
@@ -11,43 +11,204 @@
...
+                var tempHTMLFile: URL?
+                let cacheFolder: URL? = self.fileManager.getCacheDirectoryPath()
+                if cacheFolder != nil {
+                    cacheFolderURL = cacheFolder?.appendingPathComponent(self.DOWNLOAD_CACHE)
+                    tempHTMLFile = cacheFolderURL?.appendingPathComponent(self.TEMP_FILE_NAME).appendingPathExtension(self.HTML_EXTENSION)
+                    if !self.isLocalImageUsed {
+                        /* AMSDK-8942: The ACS extension saves downloaded remote image files in the cache. We have to use loadFileURL so we can allow read access to these image files in the cache but loadFileURL expects a file URL and not the string representation of the HTML payload. As a workaround, we can write the payload string to a temporary HTML file located at cachePath/adbdownloadcache/temp.html and pass that file URL to loadFileURL.
+                         */
+                        do {
+                            try FileManager.default.createDirectory(atPath: cacheFolderURL!.path, withIntermediateDirectories: true, attributes: nil)
```
nporter-adbe, line 88: Even thoes these two lines are in a do/catch, It'd be preferable to remove the usage of the force unwrapping. `cacheFolderURL!.path` and `tempHTMLFile!.path`.

AEPServices/Sources/ui/fullscreen/FullscreenMessage.swift
```diff
@@ -11,43 +11,204 @@
...
+                let keyWindow = UIApplication.shared.getKeyWindow()
+                keyWindow?.addSubview(wkWebView)
+                UIView.animate(withDuration: 0.3, animations: {
+                    var webViewFrame = wkWebView.frame
+                    webViewFrame.origin.y = 0
+                    wkWebView.frame = webViewFrame
+                }, completion: nil)
+            }
+        }
+        self.listener?.onShow()
```
nporter-adbe, line 114: Since all the above code is wrapped in an async block, the `self.listener?.onShow()` could potentially get called before the code within the block is executed. Is this intended? Just wanted to double check.

AEPServices/Sources/ui/fullscreen/FullscreenMessage.swift
```diff
@@ -11,43 +11,204 @@
...
+
+            do {
+                try FileManager.default.removeItem(atPath: tempHTMLFilePath)
+            } catch {
+                Log.debug(label: self.LOG_PREFIX, "Unable to dismiss \(error)")
+            }
+        }
+    }
+
+    // MARK: WKWebview delegatesou
```
nporter-adbe, line 143: Minor typo.

AEPServices/Sources/ui/fullscreen/FullscreenMessaging.swift
```diff
@@ -12,20 +12,18 @@
 
 import Foundation
 
-/// Fullscreen message event listener
-public protocol FullscreenListenerInterface {
-
+/// Fullscreen message lifecycle event listener
+@objc(AEPFullscreenMessaging) public protocol FullscreenMessaging {
```
nporter-adbe, line 16: Any reason we remove the `message` parameter from these functions? I feel it could be helpful.

nporter-adbe:
I know this can be a bit tough with UI code, but is there any way we can get the 16% diff test coverage up?

AEPServices/Sources/ui/GlobalUIMessaging.swift
ravjain-adb: We have this global listener as a common listener for end user so they can listen for the message monitor activities like show and dismiss. As this is a common method for all kind of message. 
I don't think message object is useful.

Also if we want to give a message back then we will need a common base class for all kind of message. But that base class might not have anything in common other than being a type for messages.

cdhoffmann (changes requested):
What happened to the floating button API? Did i miss it?

AEPServices/Sources/ui/GlobalUIMessaging.swift
```diff
@@ -0,0 +1,31 @@
+/*
+ Copyright 2020 Adobe. All rights reserved.
```
nporter-adbe, line 2: Minor, 2021.

AEPServices/Sources/ui/MessageMonitorService.swift
```diff
@@ -0,0 +1,80 @@
+/*
+ Copyright 2020 Adobe. All rights reserved.
```
nporter-adbe, line 2: 2021

AEPServices/Sources/ui/alert/AlertMessage.swift
```diff
@@ -0,0 +1,126 @@
+/*
+ Copyright 2020 Adobe. All rights reserved.
```
nporter-adbe, line 2: 2021

AEPServices/Sources/ui/alert/AlertMessaging.swift
```diff
@@ -0,0 +1,36 @@
+/*
+ Copyright 2020 Adobe. All rights reserved.
```
nporter-adbe, line 2: 2021

AEPServices/Tests/services/AlertMessageTests.swift
```diff
@@ -0,0 +1,46 @@
+/*
+ Copyright 2020 Adobe. All rights reserved.
```
nporter-adbe, line 2: 2021

AEPServices/Tests/services/FullscreenMessageTests.swift
```diff
@@ -0,0 +1,71 @@
+/*
+ Copyright 2020 Adobe. All rights reserved.
```
nporter-adbe, line 2: 2021

AEPServices/Tests/services/MessageMonitorServiceTest.swift
```diff
@@ -0,0 +1,97 @@
+/*
+ Copyright 2020 Adobe. All rights reserved.
```
nporter-adbe, line 2: 2021

nporter-adbe (commented):
To add to what Chris said, I think the pattern here should be a bit different. Instead, there should be a public protocol that defines the APIs required for a message monitor service. Then we should have a concrete class that implements these.

Even though we don't allow for overriding the service at this time, it would allow us to add that functionality down the line if needed. And makes it much easier to test as we can create a mock which conforms to the protocol.

My suggestion would be to define the APIs needed for the monitor service in something like:

```swift
protocol MessageMonitoring {
   // define your required APIs
}
```

Then implement `MessageMonitoring` in a concrete class such as: `MessageMonitorService`.
```swift
class MessageMonitorService: MessageMonitoring {
   // implement MessageMonitoring
}
```

Then in the `ServiceProvider` you can declare that the message service is of type `MessageMonitoring`.
```swift
private var defaultMessageMonitorService: MessageMonitoring = MessageMonitorService()
```

For an example of this, see the CacheService. It is declared a `Caching` type, a protocol that defines the required APIs for a Cache. Then we a concrete implementation `class DiskCacheService: Caching`. We also have some mocks of a cache, which makes protocols really useful in Swift for testing purposes.

AEPServices/Sources/ui/alert/AlertMessage.swift
```diff
@@ -0,0 +1,126 @@
...
+ OF ANY KIND, either express or implied. See the License for the specific language
+ governing permissions and limitations under the License.
+ */
+
+import Foundation
+import UIKit
+import WebKit
+
+/// Tthis class is used to create and display alert messages with custom title, message and buttons on the current view
+@objc(AEPAlertMessage)
```
nporter-adbe, line 18: Minor suggestion: One thing you could do to reduce the use of `ServiceProvider.shared.messageMonitorService` since it's a long line.

Is declare a computed variable
```swift
private var messageService: MessageMonitorService {
   return ServiceProvider.shared.messageMonitorService
}
```

Then you can just use `messageService` as a normal variable while always getting it from the service provider. You can do this in the full-screen message class too if you'd like.

AEPServices/Sources/ui/GlobalUIMessaging.swift
nporter-adbe, line 16: Since this protocol defines more of a delegate rather than a type that implements explicit functionality, I think I'd prefer to see this named something like `GlobalMessagingDelegate`.

AEPServices/Sources/ui/MessageMonitorService.swift
```diff
@@ -0,0 +1,80 @@
...
+/// This class is used to monitor if an UI message is displayed at some point in time, currently this applies for full screen and alert messages.
+/// The status is exposed through isMessageDisplayed.
+public class MessageMonitorService {
+    private let LOG_PREFIX = "MessageMonitor"
+    private var isMsgDisplayed = false
+    internal var globalUIMessagingListener: GlobalUIMessaging?
+    private let messageQueue = DispatchQueue(label: "com.adobe.uiservice.messagemonitor")
+
+    /// GlobalUIMessaging listener which is used to listen for message visibility updates.
+    public func setGlobalUIMessagingListener(listener: GlobalUIMessaging?) {
```
nporter-adbe, line 24: Currently, to set the messaging delegate we have to do:
`ServiceProvider.shared.messageMonitorService.setGlobalUIMessagingListener(listener: myListener)`
To follow a more iOS-like pattern, let's remove this setter and the variable `globalUIMessagingListener` from this class. 

Instead, add a public variable to `AEPCore`  called `messgaingDelegate` of type `GlobalMessagingDelegate`.

Then the workflow for setting a messaging delegate `AEPCore.messgaingDelegate = myDelegate`.

AEPServices/Sources/ui/MessageMonitorService.swift
```diff
@@ -0,0 +1,80 @@
...
+        self.globalUIMessagingListener = listener
+    }
+
+    /// - Returns: True if the message is being displayed else false
+    public func isMessageDisplayed() -> Bool {
```
nporter-adbe, line 29: Chatted with @sbenedicadb and it seems like this should be internal. Other extensions shouldn't care if a message is displayed. What's the use-case?

nporter-adbe (commented):
Given we move the `messagingDelegate` property to `AEPCore` and can mark `isMessageDisplayed` internal. Then we can follow our typical protocol pattern for the services, with the one exception making everything internal instead of public.

ravjain-adb:
@shalehaha @nporter-adbe  @cdhoffmann Updated the pr

shalehaha:
@ravjain-adb 
the current structure looks good to me. the only thing i don't like is the naming of `FullscreenMessaging` and `AlertMessaging`, they are listeners, but from the names it fees like they are messaging... I think we need better names

AEPServices/Sources/ui/MessageMonitorServicing.swift
```diff
@@ -12,7 +12,7 @@
 
 import Foundation
 
-/// This class is used to monitor if an UI message is displayed at some point in time, currently this applies for full screen and alert messages.
+/// This protocol is used to monitor if an UI message is displayed at some point in time, currently this applies for full screen and alert messages.
 /// The status is exposed through isMessageDisplayed.
 internal protocol MessageMonitorServicing {
```
cdhoffmann, line 17: No need for internal as the internal is the default access level.

ravjain-adb:
> AlertMessaging

We can suffix it with listener or listening ?

shalehaha:
> > AlertMessaging
> 
> We can suffix it with listener or listening ?

@nporter-adbe @cdhoffmann any suggestions here?

nporter-adbe:
If they are essentially listeners, then I would lean towards calling them delegates and referencing them as delegates in the code as well.

`FullscreenMessageDelegate` & `AlertMessageDelegate`

cdhoffmann:
I agree with @nporter-adbe . Since the delegate pattern is being used i think referencing them as delegates is the way to go.

AEPServices/Sources/ui/alert/AlertMessaging.swift
```diff
@@ -0,0 +1,40 @@
...
```
nporter-adbe, line 2: Let's make sure all the new files use 2021 as the copyright year.

nporter-adbe (approved):
Looks good pending the naming comments and CI passing.

cdhoffmann (approved):
Looks good. Just need to get the coverage up to pass CI :)

AEPServices/Sources/ui/MessagingDelegate.swift
```diff
@@ -12,20 +12,22 @@
...
-/// Global UI Message delegate which is used to listen for current message lifecycle events
-@objc(AEPGlobalUIMessaging) public protocol GlobalUIMessaging {
+/// UI Message delegate which is used to listen for current message lifecycle events
+@objc(AEPMessagingDelegate) public protocol MessagingDelegate {
 
     /// Invoked when the any message is displayed
     /// - Parameters:
     ///     - message: UIMessaging message that is being displayed
-    func onShow()
+    func onShow(message: UIMessaging?)
```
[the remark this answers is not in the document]
shalehaha: Agree with Nick, don't think it would ever be nil.
@ravjain-adb

ravjain-adb:
@cdhoffmann problem with getting coverage up is its lot of UI and we don't have viewcontroller

AEPServices/Sources/ui/fullscreen/FullscreenMessage.swift
```diff
@@ -11,43 +11,219 @@
...
+                // save the HTML payload to a local file if the cached image is being used
+                var useTempHTML = false
+                var cacheFolderURL: URL?
+                var tempHTMLFile: URL?
+                let cacheFolder: URL? = self.fileManager.getCacheDirectoryPath()
+                if cacheFolder != nil {
+                    cacheFolderURL = cacheFolder?.appendingPathComponent(self.DOWNLOAD_CACHE)
+                    tempHTMLFile = cacheFolderURL?.appendingPathComponent(self.TEMP_FILE_NAME).appendingPathExtension(self.HTML_EXTENSION)
+                    if !self.isLocalImageUsed {
+                        /* AMSDK-8942: The ACS extension saves downloaded remote image files in the cache. We have to use loadFileURL so we can allow read access to these image files in the cache but loadFileURL expects a file URL and not the string representation of the HTML payload. As a workaround, we can write the payload string to a temporary HTML file located at cachePath/adbdownloadcache/temp.html and pass that file URL to loadFileURL.
```
shalehaha, line 96: don't use internal bug numbers and it'd better not to mention `ACS extension` but make the description more general.

AEPServices/Sources/ui/fullscreen/FullscreenMessage.swift
```diff
@@ -11,43 +11,219 @@
...
+                self.webView?.removeFromSuperview()
+                self.webView = nil
+            }
+        }
     }
 
-    public func remove() {
-        fullScreenMessageHandler?.dismiss()
+    /// Open a url from this message
+    private func openUrl(url: String) {
```
shalehaha, line 223: this method is not being used?
and if we need open url then use the `URLOpening` service

ravjain-adb: Actually we don't need this function at all as we already have a openUrl service. In v5 we had that in fullscreen message.

PR #520, opened 2021-01-21
Default Header Locale Formatter

cdhoffmann:
In v5 we format the locale. Here we format the locale as in v5 except we do it in the `NetworkRequest` to keep it centrally located and not rely on each extension to do it.


AEPServices/Sources/network/DefaultHeadersFormatter.swift
```diff
@@ -0,0 +1,62 @@
...
+ */
+
+import Foundation
+
+///
+/// Helper class used to format the default http headers
+///
+public struct DefaultHeadersFormatter {
+
+    typealias httpHeaderConstants = HttpConnectionConstants.Header
```
shalehaha, line 20: do we still need this line?

AEPServices/Sources/network/DefaultHeadersFormatter.swift
```diff
@@ -0,0 +1,62 @@
...
+
+    ///
+    /// Gets the formatted locale
+    /// - Parameter unformattedLocale: The unformatted locale as a string
+    /// - Returns: The formatted locale as a string or nil if not able to format it.
+    ///
+    public static func getFormattedLocale(unformattedLocale: String) -> String {
```
shalehaha, line 27: `_ unformattedLocale`

shalehaha (approved):

AEPServices/Sources/network/DefaultHeadersFormatter.swift
cdhoffmann: Decided to change the name as well. Lmk what you think.

nporter-adbe (approved):

PR #527, opened 2021-01-27
Adding floating button to ui services

ravjain-adb:
## Description

Adding floating button with floating button delegate to listen for tap and pan gestures.

## Types of changes

## Checklist:


AEPServices/Sources/ui/floating/FloatingButtonDelegate.swift
```diff
@@ -0,0 +1,22 @@
...
+ Unless required by applicable law or agreed to in writing, software distributed under
+ the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ OF ANY KIND, either express or implied. See the License for the specific language
+ governing permissions and limitations under the License.
+ */
+
+import Foundation
+
+/// Fullscreen message lifecycle event listener
+@objc(AEPFloatingButtonDelegate) public protocol FloatingButtonDelegate {
```
cdhoffmann, line 16: Can we add an onShow and onDismiss to the delegate as well. Will be nice for testing.

cdhoffmann (commented):
Just the one comment from me. Looks good otherwise.

AEPServices/Sources/ui/UIUtils.swift
```diff
@@ -0,0 +1,32 @@
...
+ OF ANY KIND, either express or implied. See the License for the specific language
+ governing permissions and limitations under the License.
+ */
+
+import Foundation
+import UIKit
+
+class UIUtils {
+    internal static func getFrame() -> (frame: CGRect?, screenBounds: CGSize?)? {
+        var newFrame = CGRect(x: 0, y: 0, width: 0, height: 0)
```
nporter-adbe, line 18: Minor, could also write as `var newFrame = CGRect.zero`

AEPServices/Sources/ui/UIUtils.swift
```diff
@@ -0,0 +1,32 @@
...
...DLQnPgWbOP4L+L9GDgZ4X2LBmtpAzifZV2GC65HuOCA0B/D7YAi58HJcX+LFn9FwymH8XAUJh2T4BAUnn7Y9uLZrtYonpAniDwkOMLVYNLff9P4dZClyQAT9KcRyqcpa5uyZhmqAyjo3GW6PfPSa7pkjt6XP+5nGiD6IBomZyUqlikh/rAb8oVfsVxIYbdF44xijcCBRvALwZ8GGsVfcDglNMb7Ud28HxM0PsMVBJbgCaCZGM+FeCJunMtZYSv0ZSxaHUtwn1h9QfzIHLZ5E0gu+fGSK3VAg3uKfAWAoKW/v73Vr/oBNw9wTX/W0EY/F9wJOZOfbOyOv4QQXxNp5XqpvO6a94q/3zXz1smVFVlf8eV+5StfUWi18r38S8IEQ3W5gCOnfQjG+xDand13z5zru3fifyuN5lbyHdg3478dextg2wrItvL+Rmh5cJwLqrZj30Rbx+krgGgH8Yx7i27o7fCdYeqwxgka0pKmtOs03Ehimi64c33UTvNk3syjeTXPXPstg2WBVoz2LKNl5dnyx0H92J2vDP+7rMns/S6bwUEZo/Fy3pu27oLuowUAWky+FnSXLb56u8wWQGhhj9CIdn9yunD7rKHRonvE7hjCFACBfbOTkQ4g8bd7rJxRvTZj8LfLAy8zhjCeNh0edm2YNnb8o4huri3nJR/Nrk7QZOD8xHCnfm2UG6Xfu58YPxo0l6Np644LZauA4peA9nytgm60hUK1AHAtlKr+QVRQhWGUGW7D4eZ/TUujHnypQyhpM3hIjNzyXGWJW2CFYeTZhvuUqDvaKDxL9GqCK/5cC7DKjfx3W1I9l2R/osFkAbLJDZGff+JtWjaXKWyOji74X7HlUjdWAAA5UCHQvOBioQGvGFYFnqYbecc5u0wz6gcwsi6QhYEYkSRomm6f4U6lIPKCw3Hs+76Sri3A5E/d/YdyEPgpAlNdnrC6Kqz54qfF3QCaRcrlLdADFeiBkmCjwdPFF1/M/8lPS3SjEUbQzacOBI7hGLmV0EjQ8B9ESRCRRiKNAFNNK4PpLj2im0+55wpzlx4VzpX5+bynB+s0JuFlMAJG5WJEF1SY4qKLLhJY5fnnnx+AqrlcEoARsOdWAybJfTQCK9OoOZOginfeevLR8f9pdfSttJ/Wws6nXD0cbj5RIsx/N+DMVTH/H12VRhzaJ53wAAAAAElFTkSuQmCC"
```
nporter-adbe, line 31: Can we put this at the top of the file?

nporter-adbe (commented):
looks good, very minor comments.

AEPServices/Sources/ui/UIUtils.swift
ravjain-adb: Its too big and the function will move way down in the file because of this 
so kept it at the bottom

cdhoffmann (approved):

AEPServices/Sources/ui/UIUtils.swift
```diff
@@ -0,0 +1,32 @@
...
+ Unless required by applicable law or agreed to in writing, software distributed under
+ the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ OF ANY KIND, either express or implied. See the License for the specific language
+ governing permissions and limitations under the License.
+ */
+
+import Foundation
+import UIKit
+
+class UIUtils {
```
cdhoffmann, line 16: Something i am running into while writing tests for the preview manager is that any business logic which relies on the results from `UIUtils` will fail because the unit tests cannot produce a frame. We should make this class mockable in my opinion and it should either be injected into the UI elements (fullscreen message, alert, button), or it should be part of the service provider. Any thoughts on this? @shalehaha @nporter-adbe

cdhoffmann (changes requested):

AEPServices/Sources/ui/UIUtils.swift
nporter-adbe: Agreed, if the results of this class need to be used in unit tests somewhere then we should define UIUtils to conform to a protocol and injected for easy testing.

AEPServices/Sources/ui/floating/FloatingButton.swift
```diff
@@ -0,0 +1,174 @@
...
+        }
+    }
+
+    @objc private func tapDetected(recognizer: UITapGestureRecognizer) {
+        DispatchQueue.main.async {
+            self.listener?.onTapDetected()
+        }
+    }
+
+    @objc private func bringFloatingButtonToFront(timer: Timer) {
```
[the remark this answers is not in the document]
ravjain-adb: Its a selector in `Timer.scheduledTimer` at line 49. If i remove the objc it gives an error

shalehaha:
@ravjain-adb 
I just had a quick discussion with @cdhoffmann, the root problem was not the getFrame is not mocked, it is more on the Fullscreen message is not mocked. So getFrame is not something we have to mock.
One way is just to have each extension deal with this individually, they can use a factory pattern or use a wrapper to encapsulate the UI related code and make that testable.
Another solution is to make UIServices the interface to create Fullscreen/Alert message, which is also mockable.

PR #528, opened 2021-01-29
Load ECID earlier

nporter-adbe:
Not currently a bug, but to be safe we load the ECID immediately on initialization of the Identity extension.
Ref: https://jira.corp.adobe.com/browse/AMSDK-10233

### DIFF
--- a/AEPIdentity/Sources/IdentityState.swift
+++ b/AEPIdentity/Sources/IdentityState.swift
@@ -32,6 +32,7 @@ class IdentityState {
     /// - Parameter pushIdManager: a push id manager
     init(identityProperties: IdentityProperties, hitQueue: HitQueuing, pushIdManager: PushIDManageable) {
         self.identityProperties = identityProperties
+        self.identityProperties.loadFromPersistence()
         self.hitQueue = hitQueue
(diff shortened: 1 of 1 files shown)


AEPIdentity/Sources/IdentityState.swift
```diff
@@ -32,6 +32,7 @@ class IdentityState {
     /// - Parameter pushIdManager: a push id manager
     init(identityProperties: IdentityProperties, hitQueue: HitQueuing, pushIdManager: PushIDManageable) {
         self.identityProperties = identityProperties
+        self.identityProperties.loadFromPersistence()
```
emdobrin, line 35: We don't need to load it in bootupIfReady anymore.

emdobrin, line 35: Do we have e2e tests covering everything is loaded up properly?

nporter-adbe:
Given we don't believe this will be an issue in Swift and it's going to require updating tests and addressing your question over E2E tests I'm closing this for now. 

@emdobrin feel free to log an issue if you find this is a bug in the Swift SDK and we can prioritize it appropriately.

PR #531, opened 2021-02-02
Improve shared state logging and don't add empty adId to event data

nporter-adbe:
- Improve shared state logging to differentiate between standard and XDM shared state
- Don't add empty adId to IdentityProperties event data


AEPIdentity/Sources/IdentityProperties.swift
```diff
@@ -51,7 +51,9 @@ struct IdentityProperties: Codable {
     func toEventData() -> [String: Any] {
         var eventData = [String: Any]()
         eventData[IdentityConstants.EventDataKeys.VISITOR_ID_ECID] = ecid?.ecidString
-        eventData[IdentityConstants.EventDataKeys.ADVERTISING_IDENTIFIER] = advertisingIdentifier
+        if let adId = advertisingIdentifier, !adId.isEmpty {
```
kevinlind, line 54: We should add this same `isEmpty` check for pushIdentifier, blob, and locationHint.

kevinlind (approved):

PR #532, opened 2021-02-03
Simplify Zip Slip validation

cdhoffmann:
Manual canonicalization on a physical device produced unexpected results. Simply checking if the url attempts traversal via "../" is the current solution.

### DIFF
--- a/AEPServices/Sources/utility/unzip/URL+Validator.swift
+++ b/AEPServices/Sources/utility/unzip/URL+Validator.swift
@@ -18,7 +18,7 @@ extension URL {
         if self.absoluteString.contains("../") {
             return false
         }
-        
+
         return true
     }
 }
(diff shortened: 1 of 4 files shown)


AEPServices/Sources/utility/unzip/URL+Validator.swift
```diff
@@ -0,0 +1,24 @@
...
+ OF ANY KIND, either express or implied. See the License for the specific language
+ governing permissions and limitations under the License.
+ */
+
+
+extension URL {
+    ///
+    /// Validates a URL against a Zip Slip attack. Simply checks if there is any sort of traversal attempted in the url
+    /// - Returns: True if valid, false if invalid
+    func isValidUrl() -> Bool {
```
shalehaha, line 17: change the method name to be more specific

cdhoffmann: lmk if this name works.

shalehaha (approved):

PR #533, opened 2021-02-04
UIService with Showable and Dismissible  Protocols

cdhoffmann:
Adds a UIService which serves as a factory for UI elements which conform to the new Showable / Dismissible protocols for each ui element. Will allow extensions to easily mock ui element creation, show and dismissal calls.


AEPServices/Sources/ui/AEPUIService.swift
```diff
@@ -14,19 +14,19 @@ import Foundation
...
 class AEPUIService: UIService {
-    func getFrame() -> CGRect? {
-        var newFrame = CGRect.zero
-        // x is always 0
-        newFrame.origin.x = 0
-        // for fullscreen, width and height are both full screen
-        let keyWindow = UIApplication.shared.getKeyWindow()
-        guard let screenBounds: CGSize = keyWindow?.frame.size else { return nil }
-        newFrame.size = screenBounds
+    func createFullscreenMessage(payload: String, listener: FullscreenMessageDelegate?, isLocalImageUsed: Bool? = nil) -> FullscreenPresentable {
```
shalehaha, line 17: it is weird to have isLocalImageUsed as `Bool?`, we can use false as the default value

AEPServices/Sources/ui/fullscreen/FullscreenMessage.swift
```diff
@@ -45,7 +46,7 @@ public class FullscreenMessage: NSObject, WKNavigationDelegate, UIMessaging {
     /// - Parameters:
     ///     - payload: String html content to be displayed with the message
     ///     - listener: `FullscreenMessageDelegate` listener to listening the message lifecycle.
-    public convenience init(payload: String, listener: FullscreenMessageDelegate?) {
+    convenience init(payload: String, listener: FullscreenMessageDelegate?) {
```
shalehaha, line 49: I think we can get rid of the init

shalehaha (approved):

AEPServices/Sources/ui/UIService.swift
```diff
@@ -13,12 +13,34 @@
...
-    /// Background image required by the floating button
-    var ENCODED_BACKGROUND_PNG: String { get }
+    ///
+    /// Creates a `FullscreenPresentable`
+    /// - Parameters:
+    ///     - payload: The payload used in the FullscreenPresentable as a string
+    ///     - listener: The `FullscreenPresentable`'s `FullscreenMessageDelegate`
+    ///     - isLocalImageUsed: An optional flag indicating if a local image is used instead of the default image provided
+    /// - Returns: A `FullscreenPresentable` implementation
+    func createFullscreenMessage(payload: String, listener: FullscreenMessageDelegate?, isLocalImageUsed: Bool) -> FullscreenPresentable
```
ravjain-adb, line 28: As we are using the UIService for creation messages shouldn't we also inject the messagemonitor directly here and remove it from the service. As all the functions in it are internal.

shalehaha: @ravjain-adb 
I'm fine with keeping them separate, I feel like it is even cleaner today, we don't have to inject everything into one interface unless if that is necessary. What do you think?

ravjain-adb: Hmmm my only point was we purposely made the monitor service all internal because initially we did not have a factory to create all these message. Now as we are doing it I was thinking wont it be cleaner if we remove it from ServiceProvider as it is not useful for customers and its not actually a service 
What do u think ?

cdhoffmann: I can go both ways. I think it's cleaner as it is now because it's less to inject (reduces the parameters at the interface level). Technically we could adopt this pattern in the future and keep internal "services" in the service provider as internal ivars to only use internally. But if we want to keep the ServiceProvider for only public services, then i could see the argument.

shalehaha: we can remove messagemonitor from the service provider, and make it a private field of the AEPUIService and then it can be passed to fullscreen/alert message. 
@ravjain-adb is this the same with what you thought?

ravjain-adb: Yes

cdhoffmann: We will need to make the `MessageMonitoring` protocol public in this case.

shalehaha: https://git.corp.adobe.com/dms-mobile/bourbon-core-java-target/blob/f847082045a7de3cad903d90bdf118f63027ce85/code/target/src/main/java/com/adobe/marketing/mobile/TargetPreviewManager.java#L398, Target today need to access this API. Don't know whether it needs to do that, since UIService should be able to handle this logic under the hood.  @cdhoffmann do you think we need this API today?

I will create a new ticket for making this change, for now, it is an internal service, so later we can change it to any way we want. But we need a bit more info from Target/Messaging and maybe Campaign before deciding which one is the best solution.
If we need to expose a public API like `isMessageDisplayed`, I would prefer to have a separate service for it.

shalehaha: > We will need to make the `MessageMonitoring` protocol public in this case.

No, it can be hidden and part of the AEPUIService implementaion.

cdhoffmann: > > We will need to make the `MessageMonitoring` protocol public in this case.
> 
> No, it can be hidden and part of the AEPUIService implementaion.

If it is going to be injected into the `create` functions it will need to be public because those methods are part of a public protocol.

shalehaha: @cdhoffmann @ravjain-adb 
I'm also thinking we should add an `onShowFailed` delegate method to https://github.com/adobe/aepsdk-core-ios/blob/dev-v3.0.1/AEPServices/Sources/ui/fullscreen/FullscreenMessageDelegate.swift#L16 and https://github.com/adobe/aepsdk-core-ios/blob/dev-v3.0.1/AEPServices/Sources/ui/alert/AlertMessageDelegate.swift

shalehaha: > > > We will need to make the `MessageMonitoring` protocol public in this case.
> > 
> > 
> > No, it can be hidden and part of the AEPUIService implementaion.
> 
> If it is going to be injected into the `create` functions it will need to be public because those methods are part of a public protocol.

it won't be in any public interface, it can be just internal.

ravjain-adb: > > > We will need to make the `MessageMonitoring` protocol public in this case.
> > 
> > 
> > No, it can be hidden and part of the AEPUIService implementaion.
> 
> If it is going to be injected into the `create` functions it will need to be public because those methods are part of a public protocol.

Basically we don't ask the customer pass the object but when we create the message (full/alert) we pass it in the init

cdhoffmann: > > > > We will need to make the `MessageMonitoring` protocol public in this case.
> > > 
> > > 
> > > No, it can be hidden and part of the AEPUIService implementaion.
> > 
> > 
> > If it is going to be injected into the `create` functions it will need to be public because those methods are part of a public protocol.
> 
> it won't be in any public interface, it can be just internal inside AEPUIService which is the concreate class.

I thought the idea was to have the private var concrete implementation of the messageMonitor in the AEPUIService but we would be injecting that private var into the `createFullScreenMessage` and `createAlertMessage`?

cdhoffmann: > > > > We will need to make the `MessageMonitoring` protocol public in this case.
> > > 
> > > 
> > > No, it can be hidden and part of the AEPUIService implementaion.
> > 
> > 
> > If it is going to be injected into the `create` functions it will need to be public because those methods are part of a public protocol.
> 
> Basically we don't ask the customer pass the object but when we create the message (full/alert) we pass it in the init. So we don't have to make it public.

That will make testability more difficult no?

ravjain-adb: > https://git.corp.adobe.com/dms-mobile/bourbon-core-java-target/blob/f847082045a7de3cad903d90bdf118f63027ce85/code/target/src/main/java/com/adobe/marketing/mobile/TargetPreviewManager.java#L398, Target today need to access this API. Don't know whether it needs to do that, since UIService should be able to handle this logic under the hood. @cdhoffmann do you think we need this API today?
> 
> I will create a new ticket for making this change, for now, it is an internal service, so later we can change it to any way we want. But we need a bit more info from Target/Messaging and maybe Campaign before deciding which one is the best solution.
> If we need to expose a public API like `isMessageDisplayed`, I would prefer to have a separate service for it.

So extensions can listen for the message show and dismiss event and keep a flag for themselves if they want to use isMessageDisplayed

cdhoffmann: > https://git.corp.adobe.com/dms-mobile/bourbon-core-java-target/blob/f847082045a7de3cad903d90bdf118f63027ce85/code/target/src/main/java/com/adobe/marketing/mobile/TargetPreviewManager.java#L398, Target today need to access this API. Don't know whether it needs to do that, since UIService should be able to handle this logic under the hood. @cdhoffmann do you think we need this API today?
> 
> I will create a new ticket for making this change, for now, it is an internal service, so later we can change it to any way we want. But we need a bit more info from Target/Messaging and maybe Campaign before deciding which one is the best solution.
> If we need to expose a public API like `isMessageDisplayed`, I would prefer to have a separate service for it.

My understanding is that the extensions no longer need this functionality because the UI elements (fullscreen message and alert) use the message monitor to know check if something is displayed or not. Basically the logic was moved into the ui layer.

shalehaha: > > > > > We will need to make the `MessageMonitoring` protocol public in this case.
> > > > 
> > > > 
> > > > No, it can be hidden and part of the AEPUIService implementaion.
> > > 
> > > 
> > > If it is going to be injected into the `create` functions it will need to be public because those methods are part of a public protocol.
> > 
> > 
> > Basically we don't ask the customer pass the object but when we create the message (full/alert) we pass it in the init. So we don't have to make it public.
> 
> That will make testability more difficult no?

we won't change your current API in UIServices, so it won't impact testability.

Here are the changes I think we can make based on the current in this PR.
1. Remove message monitor from Service provider, change it to an instance variable of AEPUiServcies
2. Change the init methods of alert and fullscreen message to accept a message monitor object, and use that instead of getting it from the Service provider.
3. Add an `onShowFailed` delegate method to alert and fullscreen delegate interface, it should be called whenever we failed to show a messaging.

shalehaha: > > https://git.corp.adobe.com/dms-mobile/bourbon-core-java-target/blob/f847082045a7de3cad903d90bdf118f63027ce85/code/target/src/main/java/com/adobe/marketing/mobile/TargetPreviewManager.java#L398, Target today need to access this API. Don't know whether it needs to do that, since UIService should be able to handle this logic under the hood. @cdhoffmann do you think we need this API today?
> > I will create a new ticket for making this change, for now, it is an internal service, so later we can change it to any way we want. But we need a bit more info from Target/Messaging and maybe Campaign before deciding which one is the best solution.
> > If we need to expose a public API like `isMessageDisplayed`, I would prefer to have a separate service for it.
> 
> My understanding is that the extensions no longer need this functionality because the UI elements (fullscreen message and alert) use the message monitor to know check if something is displayed or not. Basically the logic was moved into the ui layer.

we won't change your current API in UIServices, so it won't impact testability.
Here are the changes I think we can make based on the current in this PR.
Remove message monitor from Service provider, change it to an instance variable of AEPUiServcies
Change the init methods of alert and fullscreen message to accept a message monitor object, and use that instead of getting it from the Service provider.
Add an onShowFailed delegate method to alert and fullscreen delegate interface, it should be called whenever we failed to show a messaging.

ravjain-adb: We can still mock messageMonitor and pass it to messages right ? Wont it be useful while testing full and alert message?

cdhoffmann: Yup yup. It would only affect "testing" the MessageMonitor implementation itself which we don't need to do obviously. I was thinking of it backwards. 👍

shalehaha: we can still keep it as a protocol if that helps with unit tests.
I'm merging this PR, @cdhoffmann you can make those new changes on a new one.

PR #535, opened 2021-02-04
Remove message monitor from service provider, add onShowFailed to delegates

cdhoffmann:
Now that we are using a UIService to provide the UI elements, we no longer need the MessagingMonitor to be a service.

### DIFF
--- a/AEPServices/Sources/ui/alert/AlertMessageDelegate.swift
+++ b/AEPServices/Sources/ui/alert/AlertMessageDelegate.swift
@@ -41,5 +41,5 @@ import Foundation
     ///
     /// Invoked when the alert failed to be displayed
     ///
-    func onShowFailed()
+    func onShowFailure()
 }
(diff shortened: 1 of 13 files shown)


AEPServices/Sources/ui/alert/AlertMessageDelegate.swift
```diff
@@ -37,4 +37,9 @@ import Foundation
     ///     - message: Alert message which is currently dismissed
     @objc(onDismissWithAlertMessage:)
     func onDismiss(message: AlertMessage)
+    
+    ///
+    /// Invoked when the alert failed to be displayed
+    ///
+    func onShowFailed()
```
shalehaha, line 44: is it better to rename it to `onFailure` or `onShowFailure`?

shalehaha (approved):

ravjain-adb (approved):
LGTM

Issue #519, opened 2021-01-20
Strings not logged correctly


nporter-adbe:
When printing shared state and other event data to the logging service it appears that occasionally quotation marks are left out when printing these values. I double-checked in the debugger and `mid` is a `String`, however, it prints as if it's an `Int`.

```
2021-01-19 14:13:31.943096-0800 School App AEP[13386:362144] [AEP SDK DEBUG - <EventHub>] Shared state created for com.adobe.module.identity with version 0 and data: 
{
  blob = j8Odv6LonN4r3an7LhD3WZrU1bUpAkFkkiY1ncBR96t2PTI;
  lastsync = "1611082533.758741";
  locationhint = 9;
  mid = 65837975888089377005040020619800013833;
}
```

yangyansong-adbe:
This is as expected in Swift, since we're printing `event data` as an `AnyObject`.
https://github.com/adobe/aepsdk-core-ios/blob/07423d952bef7bacaaac98ea1abc0e76e783107a/AEPCore/Sources/eventhub/EventHub.swift#L184

See my sample code and output => 

```Swift
var dict = [String: Any]()
dict = [
    "a": "13435454",
    "b":[
        "b1": 1235566,
        "b2": "35566"
    ]
]
print(dict as AnyObject)
print(dict)
```

```
{
    a = 13435454;
    b =     {
        b1 = 1235566;
        b2 = 35566;
    };
}
["a": "13435454", "b": ["b2": "35566", "b1": 1235566]]
```

yangyansong-adbe:
@nporter-adbe I think this is not a bug, can we close it?

nporter-adbe:
It may not be a bug exactly, but the logs are misleading. The reason I originally opened this issue is that someone asked why we were sharing the mid as a number in Swift. In the original log, I posted it looks like the mid is a number and not a string. 

I think it'd be best if the logs accurately portrayed the data types of the values being printed. My suggestion would be to potentially serialize the event data as pretty-printed JSON, then print it. If serialization fails, just print it as `Any`.

PR #536, opened 2021-02-05
print EventData as a prettified JSON string (option 1)

yangyansong-adbe:
Adds a `string interpolation` function to convert EventData to a prettified JSON string. And I have to make the extension public since it will be used in other modules.

### DIFF
--- a/AEPServices/Sources/log/DefaultStringInterpolation+EventDataPrinter.swift
+++ b/AEPServices/Sources/log/DefaultStringInterpolation+EventDataPrinter.swift
@@ -3,21 +3,23 @@
  This file is licensed to you under the Apache License, Version 2.0 (the "License");
  you may not use this file except in compliance with the License. You may obtain a copy
  of the License at http://www.apache.org/licenses/LICENSE-2.0
+
  Unless required by applicable law or agreed to in writing, software distributed under
  the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
  OF ANY KIND, either express or implied. See the License for the specific language
  governing permissions and limitations under the License.
  */
 
-extension URL {
-    ///
-    /// Validates a URL against a Zip Slip attack. Simply checks if there is any sort of traversal attempted in the url
-    /// - Returns: True if valid, false if invalid
-    func isSafeUrl() -> Bool {
-        if self.absoluteString.contains("../") {
-            return false
-        }
+import Foundation
 
-        return true
+public extension DefaultStringInterpolation {
+    mutating func appendInterpolation(prettyEventData value: [String: Any]) {
+        guard JSONSerialization.isValidJSONObject(value),
+              let data = try? JSONSerialization.data(withJSONObject: value, options: .prettyPrinted),
+              let prettyPrintedString = NSString(data: data, encoding: String.Encoding.utf8.rawValue) else {
+            appendInterpolation(value as AnyObject)
+            return
+        }
+        appendInterpolation(prettyPrintedString)
     }
 }
--- a/AEPServices/Sources/log/DefaultStringInterpolation+EventDataPrinter.swift
+++ b/AEPServices/Sources/log/DefaultStringInterpolation+EventDataPrinter.swift
@@ -16,7 +16,7 @@ public extension DefaultStringInterpolation {
     mutating func appendInterpolation(prettyEventData value: [String: Any]) {
         guard JSONSerialization.isValidJSONObject(value),
               let data = try? JSONSerialization.data(withJSONObject: value, options: .prettyPrinted),
-              let prettyPrintedString = NSString(data: data, encoding: String.Encoding.utf8.rawValue) else {
+              let prettyPrintedString = String(data: data, encoding: String.Encoding.utf8) else {
             appendInterpolation(value as AnyObject)
             return
         }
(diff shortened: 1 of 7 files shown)


AEPServices/Sources/log/DefaultStringInterpolation+EventDataPrinter.swift
```diff
@@ -0,0 +1,25 @@
...
+
+ Unless required by applicable law or agreed to in writing, software distributed under
+ the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ OF ANY KIND, either express or implied. See the License for the specific language
+ governing permissions and limitations under the License.
+ */
+
+import Foundation
+
+public extension DefaultStringInterpolation {
```
nporter-adbe, line 15: Looks good, I'm a bit confused where this is used by the logging service? Also, I'd prefer if this wasn't a public extension so we aren't exposing to customers if possible.

nporter-adbe: Appears that this affects how all interpolated strings are printed in swift, even outside the SDK. I'm not sure we can go this route as it could potentially affect customer's logging and other string manipulations.

yangyansong-adbe: The logging service just simply accepts any string as an argument and print it, so what we need to do is to provide a prettified `Event Data` JSON string by using either a utility function or in this PR using a `string interpolation` function .
In fact, the approach of enhancing `string interpolation` do have the side effect which affects customer's String manipulations. But the customer will not run into this code if they don't use it in their code.
```Swift
"\(prettyEventData:["key":"value"])"
```
Either way is ok with me, @nporter-adbe  @shalehaha @pfransenadb any comments/ideas?

nporter-adbe: Can we make this extension internal so it won't affect customers in any way?

yangyansong-adbe: We can't, if so, other AEP extensions will not have access to `prettyEventData `.

nporter-adbe: It's not clear to me how this is intended to be used. Does the logging service use it by default or do extensions need to make changes to use this? Can you provide sample usage in the PR description?

nporter-adbe:
This provides a pathway to fixing the issue that was opened in #519, but doesn't fix it. In this PR I think we also need to update the EventHub to print all data pretty printed.

nporter-adbe:
I'm ok with it being public but I think we need a minor change:
1. Let's update the name (prettyEventData) to have the `AEP` prefix so it's clear it comes from our SDK. 
2. Let's add documentation about this new functionality: https://github.com/adobe/aepsdk-core-ios/blob/main/Documentation/Services/Logging.md
3. Update event data printing in EventHub.

@shalehaha what do you think about it being public?

shalehaha:
My main concern is on the performance side. Now we need to convert a dict to a JSON then to a string. And the log filter is only applied after the string interpolation, so that will be a waste of resources.
We should run some performance test, and explorer ways to only run string interpolation code when the log level fits.

yangyansong-adbe:
> My main concern is on the performance side. Now we need to convert a dict to a JSON then to a string. And the log filter is only applied after the string interpolation, so that will be a waste of resources.


@shalehaha check the following test result, it shows that the default string interpolation function which handles `\(data as AnyObject)` used more time than custom interpolation function. I guess the reason is because the default interpolation function uses some generic algorithms.

> We should run some performance test, and explore ways to only run string interpolation code when the log level fits.

@shalehaha I have no idea how to do it, any options/thoughts? 

```Swift 
func testPerformance_1(){
        var innerData = [String: Any]()
        for n in 0..<100{
            innerData["A_\(n)"] = "v_\(n)"
        }
        let event = Event(name: "test event", type: "test_type", source: "test_source", data:
                            [
                                "k1":"v2",
                                "k2":[
                                    "x1":"v",
                                    "x2":"v",
                                    "x3":[
                                        "y1":"v",
                                        "y2":"v",
                                        "y3":innerData
                                    ]
                                ]
                            ])
        guard let data = event.data else {return}
        self.measure {
            for _ in 0..<10000{
                Log.debug(label: "ptest", "==== \(data as AnyObject)")
            }
        }
    }
    func testPerformance_2(){
        var innerData = [String: Any]()
        for n in 0..<100{
            innerData["A_\(n)"] = "v_\(n)"
        }
        let event = Event(name: "test event", type: "test_type", source: "test_source", data:
                            [
                                "k1":"v2",
                                "k2":[
                                    "x1":"v",
                                    "x2":"v",
                                    "x3":[
                                        "y1":"v",
                                        "y2":"v",
                                        "y3":innerData
                                    ]
                                ]
                            ])
        guard let data = event.data else {return}
        self.measure {
            for _ in 0..<10000{
                Log.debug(label: "ptest", "==== \(prettyEventData: data)")
            }
        }
    }
```

output: 
```Swift
Test Case '-[AEPCoreTests.EventDataFlattenerTests testPerformance_1]' started.
measured [Time, seconds] average: 2.160, relative standard deviation: 6.042%, values: [2.017680, 2.388317, 2.141059, 2.174101, 2.018301, 2.294322, 2.254293, 2.261874, 2.061487, 1.986043], performanceMetricID:com.apple.XCTPerformanceMetric_WallClockTime, baselineName: "", baselineAverage: , maxPercentRegression: 10.000%, maxPercentRelativeStandardDeviation: 10.000%, maxRegression: 0.100, maxStandardDeviation: 0.100
Test Case '-[AEPCoreTests.EventDataFlattenerTests testPerformance_1]' passed (21.998 seconds).
Test Case '-[AEPCoreTests.EventDataFlattenerTests testPerformance_2]' started.
measured [Time, seconds] average: 1.639, relative standard deviation: 5.874%, values: [1.544388, 1.507246, 1.495523, 1.735524, 1.763290, 1.598597, 1.679254, 1.612203, 1.694888, 1.756876], performanceMetricID:com.apple.XCTPerformanceMetric_WallClockTime, baselineName: "", baselineAverage: , maxPercentRegression: 10.000%, maxPercentRelativeStandardDeviation: 10.000%, maxRegression: 0.100, maxStandardDeviation: 0.100
Test Case '-[AEPCoreTests.EventDataFlattenerTests testPerformance_2]' passed (16.644 seconds).
```

shalehaha:
I think using a utility class is more flexible and won't pollute the customer's namespace.

yangyansong-adbe:
@nporter-adbe  @shalehaha  https://github.com/adobe/aepsdk-core-ios/pull/562 in the (option 2) PR, the event data is printed using a utility class. If we're leaning towards option 2, I will close this one (option 1).

Issue #433, opened 2020-10-28
visitoridslist has [AEPIdentity.CustomIdentity] type in Identity shared state


emdobrin:
### Expected Behaviour
visitoridslist should have raw data format when sharing through Shared States

### Actual Behaviour
visitoridslist has [AEPIdentity.CustomIdentity] type in Identity shared state

### Reproduce Scenario (including but not limited to)
Using Swift core and a free form Data element in Launch for `~state.com.adobe.module.identity/visitoridslist` I noticed the returned value from the client is [AEPIdentity.CustomIdentity]

emdobrin:
I updated my app to AEPIdentity v3.0.0-beta.1 and I observed the following shared state:
```
"visitoridslist": [Optional(["authentication.state": 0, "id": dc565042aa48e653af4d84a3ba8043757112006556aac9560084a1243e5fb773, "id.type": testEmail, "id.origin": d_cid_ic])]
```
The visitor ids lists in ACPCore were id, id_origin, id_type and authentication_state, so we will need to change the keys in AEPIdentity to match for backwards compatibility.

nporter-adbe:
@emdobrin thanks for bringing this up, I checked the ACPCore code and it matches with the keys you have listed. Last week @cdhoffmann and I validated some stuff in Griffon and saw the visitor id list to look a bit different in V5. Here is a sample event from V5 in Griffon

Any idea why this is different? This is definitely from V5 as the event number is populated.

```
{
  "ACPExtensionEventUniqueIdentifier": "987D9364-423F-47AF-83BF-4BFD54FB6F97",
  "ACPExtensionEventNumber": 30,
  "ACPExtensionEventData": {
    "visitoridslist": [
      {
        "id": "1234567",
        "authentication.state": 1,
        "id.origin": "d_cid_ic",
        "id.type": "idType1"
      }
    ],
    "lastsync": 1603299546,
    "mid": "47638672570484457245556507411373198888",
    "blob": "j8Odv6LonN4r3an7LhD3WZrU1bUpAkFkkiY1ncBR96t2PTI",
    "locationhint": "9"
  },
  "ACPExtensionEventSource": "com.adobe.eventsource.responseidentity",
  "ACPExtensionEventName": "UPDATED_IDENTITY_RESPONSE",
  "ACPExtensionEventType": "com.adobe.eventtype.identity"
}
```

nporter-adbe:
[...]

PR #550, opened 2021-02-25
Change CustomIdentity to use underscores during encoding

nporter-adbe:
https://github.com/adobe/aepsdk-core-ios/issues/433


emdobrin:
We should have tests validating the format.
Secondly, what happens with previously migrated ids, will those be lost in 3.0.0?

nporter-adbe:
@emdobrin that's a good point about the migration aspect, I'm not sure if changing the coding keys will have an impact. I can run a quick test after PTO.

nporter-adbe:
@emdobrin So I don't believe we need to worry about migrating. The initializer for the CustomIdentity wasn't accessible publicly, so I don't believe it would have been possible to go through this workflow.

nporter-adbe:
Will be adding tests shortly.

nporter-adbe:
@emdobrin, I did confirm that changing the coding keys will break loading it from persistence. Given that I don't believe there was any way to set `CustomIdentity` in the SDK, given the init level was internal, I'm not sure we need to worry about that here.

AEPIdentity/Sources/CustomIdentity.swift
```diff
@@ -24,19 +24,46 @@ public class CustomIdentity: Identifiable, Codable {
...
+        do {
+            // Attempt to decode using the correct keys
+            let values = try decoder.container(keyedBy: CodingKeys.self)
+            origin = try? values.decode(String.self, forKey: .origin)
+            type = try? values.decode(String.self, forKey: .type)
+            identifier = try? values.decode(String.self, forKey: .identifier)
+            authenticationState = try values.decode(MobileVisitorAuthenticationState.self, forKey: .authenticationState)
+        } catch _ {
+            // Decoding failed using correct keys, attempt to decode using migration keys
+            let values = try decoder.container(keyedBy: MigrationCodingKeys.self)
```
nporter-adbe, line 60: If decoding fails we attempt to use the old "Migration" keys. @emdobrin

emdobrin: I think we should either include this in the migration class or expect a new sync call from the developer for the lost ids.  @shalehaha what do you think? this manifests if migrating from v4/v5->3.0.0.

AEPIdentity/Sources/CustomIdentity.swift
```diff
@@ -24,18 +24,18 @@ public class CustomIdentity: Identifiable, Codable {
     ///   - type: Type of the identifier
     ///   - identifier: The identifier
     ///   - authenticationState: Authentication state for the identifier
-    init(origin: String?, type: String?, identifier: String?, authenticationState: MobileVisitorAuthenticationState) {
+    public init(origin: String?, type: String?, identifier: String?, authenticationState: MobileVisitorAuthenticationState) {
```
emdobrin, line 27: Why do we need this class to be public? Looking at the Identity public APIs it looks like the sync apis accept free form dictionaries, while the getter returns Identifiable.

nporter-adbe: fixed.

emdobrin: can we remove the modifier from the class too? cc @shalehaha

nporter-adbe: removed that too. the original intention was for this class to be public, but that has been replaced with the `Identifiable` protocol.

shalehaha (approved):

AEPIdentity/Sources/CustomIdentity.swift
shalehaha: yep, should change it to private

emdobrin (approved):

PR #552, opened 2021-02-26
Regular listeners shouldn't be called for paired response events

### LINKED ISSUE #551: Regular listeners shouldn't receive paired events
### Expected Behaviour
Regular listeners should only receive the events they registered for and paired events should only be returned to response listeners. All events should be received by wildcard listeners.
Note: This is also the behavior in the ACP SDK.

### Actual Behaviour
Paired and non-paired response events are currently received by regular listeners.

emdobrin:
Fixes #551 and #557


AEPCore/Sources/eventhub/EventListenerContainer.swift
```diff
@@ -62,4 +62,8 @@ internal extension EventListenerContainer {
     init(listener: @escaping EventResponseListener, triggerEventId: UUID, timeout: DispatchWorkItem?) {
         self.init(listener: listener, type: nil, source: nil, triggerEventId: triggerEventId, timeoutTask: timeout)
     }
+
+    func isWildcard() -> Bool {
```
nporter-adbe, line 66: I'd prefer to see this as a computed variable as it describes this type.
```swift
var isWildcard: Bool { 
   return source == EventSource.wildcard && type == EventType.wildcard
}
```

AEPCore/Sources/eventhub/EventListenerContainer.swift
```diff
@@ -44,8 +44,8 @@ struct EventListenerContainer: Equatable {
     /// - Parameter event: An `Event` being dispatched by the `EventHub`
     /// - Returns: True if `listener` should be notified of `event`
     func shouldNotify(_ event: Event) -> Bool {
-        if let listenerTriggerId = triggerEventId {
-            return listenerTriggerId == event.responseID
+        if event.responseID != nil {
+            return event.responseID == triggerEventId || self.isWildcard()
         }
 
         return (event.type == type || type == EventType.wildcard)
```
nporter-adbe, line 51: Can we update the last return statement to use the `isWildcard` member?
`return (event.type == type && event.source == source) || isWildcard`

nporter-adbe (approved):
nice 👍🏻

AEPCore/Sources/eventhub/EventListenerContainer.swift
emdobrin: not really, because the current condition also covers the use-case where only one (type or source) is wildcard.

nporter-adbe:
One concern I have with this change is how response events are used in other places in the SDK places.

Currently, in configuration, we dispatch a response event when the config is updated. No specific listeners are listening for the "response" event in particular. But this is still a response event to a request event, so we want to tie them together by making it a response event. This is useful in Griffon mainly and debugging.

With this change, it would appear that no other extension could actually listen for the response event coming from configuration, as no one is registered to be a response listener on the configuration request event.

On second thought, I actually think the existing behavior may be desired. I think other listeners that match the type and source should be notified of this event. But response listeners should only be invoked when the corresponding response event comes through the EventHub and ensures it is only invoked once.

https://github.com/adobe/aepsdk-core-ios/blob/main/AEPCore/Sources/configuration/Configuration.swift#L199

@shalehaha thoughts?

nporter-adbe:
Another example I can give is that Griffon listens for Analytics response events to tie them to Analytics request events. With this change, Griffon wouldn't be able to hear the response events as those events would only be dispatched to the paired listeners and not to every listener registered for Analytics response events.

emdobrin:
In the past we used to only send a paired response event (with responseId) when there was a getter waiting for it and that was the only listener that should have receive this event, for other updates it should a generic, non-paired response event that will be broadcasted to all the listeners interested in the event. Did this behavior change in the case of configuration 3.0.0?

emdobrin:
Griffon uses wildcard listener so the response event will be visible to it.

emdobrin:
@nporter-adbe @shalehaha As discussed offline, this PR makes the regular vs paired response events behavior to be in line with the ACPCore logic. 
I updated my PR to include both the Core changes and the necessary configuration updates for dispatching paired responses for request events / getters and regular non-paired responses for all other config updates so all extensions can receive them. @nporter-adbe  thanks for the good catch.

shalehaha (approved):

PR #562, opened 2021-03-04
print EventData as a prettified JSON string (option 2)

yangyansong-adbe:


AEPServices/Sources/utility/EventDataPrinter.swift
```diff
@@ -0,0 +1,31 @@
...
+ of the License at http://www.apache.org/licenses/LICENSE-2.0
+ Unless required by applicable law or agreed to in writing, software distributed under
+ the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ OF ANY KIND, either express or implied. See the License for the specific language
+ governing permissions and limitations under the License.
+ */
+
+import Foundation
+
+public enum EventDataPrinter {
```
shalehaha, line 14: we can rename it to something like `DictionaryPretty`, since we can also use it for other type of data

shalehaha (approved):

PR #567, opened 2021-03-10
Fix IdentityProperties v4/v5 migration format

kevinlind:
## Description

Fix the format of the migrated v4 and v5 Identity properties in order to be decoded as an IdentityProperties object.

- Stores the migrated data as an encoded AnyCodable dictionary.
- Updates IDParser to match CustomIdentity coding keys.
- Fix migration datastore keys to match those used by Identity.
- Add unit tests to verify migrated identities are decodable to IdentityProperties object.

## Types of changes

## Checklist:

### DIFF
--- a/AEPCore/Sources/migration/IDParser.swift
+++ b/AEPCore/Sources/migration/IDParser.swift
@@ -38,6 +38,8 @@ struct IDParser: IDParsing {
             if currentCustomerIdOrigin.isEmpty || currentCustomerIdValue.isEmpty {
                 continue
             }
+            
+            let originValue = currentCustomerIdOrigin.components(separatedBy: "=")[0]
 
             let idValueComponents = currentCustomerIdValue.components(separatedBy: CID_DELIMITER)
 
@@ -46,7 +48,7 @@ struct IDParser: IDParsing {
                 continue
             }
 
-            let IdDict = ["origin": currentCustomerIdOrigin, "type": idValueComponents[0], "id": idValueComponents[1], "authenticationState": Int(idValueComponents[2]) ?? 0] as [String: Any]
+            let IdDict = ["id_origin": originValue, "id_type": idValueComponents[0], "id": idValueComponents[1], "authentication_state": Int(idValueComponents[2]) ?? 0] as [String: Any]
             ids.append(IdDict)
         }
 
--- a/AEPCore/Sources/migration/IDParser.swift
+++ b/AEPCore/Sources/migration/IDParser.swift
@@ -38,7 +38,7 @@ struct IDParser: IDParsing {
             if currentCustomerIdOrigin.isEmpty || currentCustomerIdValue.isEmpty {
                 continue
             }
-            
+
             let originValue = currentCustomerIdOrigin.components(separatedBy: "=")[0]
 
             let idValueComponents = currentCustomerIdValue.components(separatedBy: CID_DELIMITER)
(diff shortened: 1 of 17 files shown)


AEPCore/Sources/migration/V4MigrationConstants.swift
```diff
@@ -50,8 +50,8 @@ enum V4MigrationConstants {
         static let V4_PUSH_TOKEN = "ADBMOBILE_KEY_PUSH_TOKEN"
 
         enum DataStoreKeys {
-            static let IDENTITY_PROPERTIES = "identitiesproperties"
-            static let PUSH_ENABLED = "ADOBEMOBILE_PUSH_ENABLED"
+            static let IDENTITY_PROPERTIES = "identity.properties"
```
nporter-adbe, line 53: Looks like we forgot to update these once we decided on a format in the style guide for data store keys.

@shalehaha we may need to update the other keys as well.

AEPCore/Sources/migration/V4Migrator.swift
```diff
@@ -122,7 +122,7 @@ struct V4Migrator {
 
         // save values
         let identityDataStore = NamedCollectionDataStore(name: V4MigrationConstants.Identity.DATASTORE_NAME)
-        let identityPropsData = try? JSONSerialization.data(withJSONObject: identityPropsDict)
+        let identityPropsData = AnyCodable.from(dictionary: identityPropsDict as [String : Any])
```
nporter-adbe, line 125: Is this required? From my understanding putting the data representation of the dict should be nearly equivalent.

AEPCore/Sources/migration/V5Migrator.swift
```diff
@@ -115,7 +115,7 @@ struct V5Migrator {
 
         // save values
         let identityDataStore = NamedCollectionDataStore(name: V5MigrationConstants.Identity.DATASTORE_NAME)
-        let identityPropsData = try? JSONSerialization.data(withJSONObject: identityPropsDict)
```
nporter-adbe, line 118: Same question here.

AEPCore/Sources/migration/IDParser.swift
```diff
@@ -46,7 +48,7 @@ struct IDParser: IDParsing {
                 continue
             }
 
-            let IdDict = ["origin": currentCustomerIdOrigin, "type": idValueComponents[0], "id": idValueComponents[1], "authenticationState": Int(idValueComponents[2]) ?? 0] as [String: Any]
+            let IdDict = ["id_origin": originValue, "id_type": idValueComponents[0], "id": idValueComponents[1], "authentication_state": Int(idValueComponents[2]) ?? 0] as [String: Any]
```
nporter-adbe, line 51: Looks like these didn't get updated either when the keys were changed.

I think I'd prefer to see them something like: `[CustomIdentity.CodingKeys.origin :originValue]`

nporter-adbe: Actually, I guess that may not be possible as CustomIdentity is in AEPIdentity target.

kevinlind: Yes. It would cause a circular dependency between AEPCore and AEPIdentity, unless we moved CustomIdentity to AEPCore. I don't feel it is necessary to do that for this case, however.

AEPCore/Sources/migration/V4Migrator.swift
nporter-adbe: Gotcha, so this is an optimization. 👍🏻

nporter-adbe (approved):

PR #573, opened 2021-03-12
Add edge prefix to eventTypes

emdobrin:
As per discussion in AMSDK-11139


nporter-adbe (approved):

emdobrin:
cc @pfransenadb

AEPCore/Sources/eventhub/EventType.swift
```diff
@@ -41,6 +41,6 @@ public class EventType: NSObject {
     public static let genericPii = "com.adobe.eventType.generic.pii"
     public static let genericData = "com.adobe.eventType.generic.data"
     public static let edge = "com.adobe.eventType.edge"
-    public static let consent = "com.adobe.eventType.consent"
+    public static let edgeConsent = "com.adobe.eventType.edgeConsent"
```
emdobrin, line 44: I would prefer to keep this consent, but I also feel that if we do that it will not be consistent with the shared state name proposals and module name.

PravinPK (approved):
+1 for edgeConsent 
and consistency of having the Prefix "edge" everywhere

kevinlind (approved):

PR #583, opened 2021-03-24
Support legacy 3rd party extension registration

nporter-adbe:
## Motivation:

The current workflow when using the compat layer can be confusing, especially without any documentation. To use Swift + a legacy 3rd party extension, you must do one of the following to register extensions.

Current:
```swift
AEPAssurance.registerExtension()
MobileCore.registerExtensions(extens)
ACPCore.start {}

// or

ACPCore.registerExtensions([AEPAssurance.self, Identity.sef, etc...])
```

We want to reach the point where we can treat legacy 3rd party extensions on a level playing field with the latest Swift extensions. To do this, we want to allow extension registration of legacy 3rd parties in-line with Swift extensions.

```swift
MobileCore.registerExtensions([AEPAssurance.self, Lifecycle.self, Identity.self])
```

This PR, along with https://github.com/adobe/aepsdk-compatibility-ios/pull/23, adds support for this type of extension registration. 

## Solution

Now that we support registering a single extension and we are open to making minor changes in Core, we can support this workflow quite easily.

- We allow our register extensions API to take in an array of `NSObject.Type` instead of `Extension.Type`
- We determine if an extension is a native Swift extension or if it could be a legacy 3rd party during registration. Based on this check, we either register it as normal, or if it is a 3rd party extension, we attempt to invoke the `registerExtension` API.
- Invoking the `registerExtension` API will, in turn, use the compatibility layer to register it individually. 

## Caveats

- All extensions now must inherit from `NSObject`. This isn't an issue today in production as all our extensions are required to inherit from `NSObject` for Obj-C compatibility. However, this may break some of our mock extensions in other extensions causing tests not to compile. I don't anticipate this having breaking changes in production unless someone wrote their own Swift extension and it doesn't inherit from `NSObject`.
- I noticed that if I do not mark `@objc` to the `registerExtension` API in the 3rd party extension, then `responds("registerExtension")` returns false. We will need to communicate with partners with legacy 3rd party extensions written in Swift that they need to expose `registerExtension` to Obj-C via the `@objc` attribute.
- The compact layer will need to be present in the app for this to actually work. I'm not sure the best way to enforce this.

## Testing

I tested this code along with the update in the compact layer in the sample app. I was able to connect to Assurance without issue and see events populating in the UI, and also successfully registered the apps flyer extension.

```swift
MobileCore.registerExtensions([AppsFlyerAdobeExtension.self, AEPAssurance.self, Lifecycle.self, ...])
```

Assurance session: https://experience.adobe.com/#/@aemonacpprodcampaign/griffon/session/a59bf143-ce86-48dd-8289-3ab1423803de/view/event-list

## FAQ

> Can this code live in the compat layer?

I attempted to expose this API via an `extension` on `MobileCore` in the compat layer, but it appears our minimum version of iOS 10.0 is holding us back.

error

> Why can't we overload the API instead of modifying it?

This was the original plan, however, if we have two APIs defined as
```swift
public static func registerExtensions(_ extensions: [Extension.Type], _ completion: (() -> Void)? = nil)
public static func registerExtensions(_ extensions: [NSObject.Type], _ completion: (() -> Void)? = nil)

MobileCore.registerExtensions([Identity.self, Lifecycle.self]) // error, ambiguous use
```

Ironically, adding a new API would be a breaking change. The compiler gets confused because technically, the list of extensions we have provided is an array of `NSObject.Type` and an array of `Extension.Type`, so it doesn't know which API to use. So this would actually be a breaking change.


shalehaha:
@nporter-adbe Thanks for proposing this change, however, I want to hold off the change until we have made a decision on whether we want to provide the compatibility layer for 3rd party extension.

nporter-adbe:
@shalehaha,  @pfransenadb and @hunterpeterson have requested this change.

AEPCore/Sources/core/MobileCore.swift
```diff
@@ -41,17 +41,31 @@ public final class MobileCore: NSObject {
...
         V5Migrator(idParser: idParser).migrate() // before starting SDK, migrate from v5 if needed
 
+        // Invoke registerExtension on legacy extensions
+        let legacyExtensions = extensions.filter {!($0.self is Extension.Type)} // All extensions that do not conform to `Extension`
+        let registerSelector = Selector(("registerExtension"))
+
+        for legacyExtension in legacyExtensions
+            where legacyExtension.responds(to: registerSelector) {
+            legacyExtension.perform(registerSelector)
+            if NSClassFromString("ACPBridgeExtension") == nil {
```
pfransenadb, line 56: Should we even perform the registerSelector if we don't have the compat layer installed?  Seems like it'd be better to put the if outside of the for loop and not even run the for loop if it's not there...

pfransenadb (changes requested):
Question about bridge...

nporter-adbe:
@pfransenadb thanks, updated.

pfransenadb (approved):

nporter-adbe:
@shalehaha any concern with merging this?

Issue #563, opened 2021-03-04
getSharedState() will not return the latest if the given event is nil


yangyansong-adbe:
### Expected Behaviour

```swift
    /// Retrieves the `SharedState` for a specific extension
    /// - Parameters:
    ///   - extensionName: An extension name whose `SharedState` will be returned
    ///   - event: If not nil, will retrieve the `SharedState` that corresponds with this event's version. If nil will return the latest `SharedState`
    ///   - barrier: If true, the `EventHub` will only return `.set` if `extensionName` has moved past `event`
    ///   - sharedStateType: The type of shared state to be read from, if not provided defaults to `.standard`
    /// - Returns: The `SharedState` data and status for the extension with `extensionName`
```
From the above public API doc, getSharedState() should return the latest `SharedState` if the given even is nil.

### Actual Behaviour

https://github.com/adobe/aepsdk-core-ios/blob/9fe655414c1e88d4252e9be25cf4f8d1dc539bee/AEPCore/Sources/eventhub/EventHub.swift#L233

SharedState.resolve() will return the oldest one if the event version is 0.

nporter-adbe:
Are we sure we want to make this change and not update the doc comment?
https://github.com/adobe/aepsdk-core-ios/pull/299

nporter-adbe:
Talked with @shalehaha the current implementation is desired, however, somehow the doc comment got out of sync. To resolve this issue I think we just need to update the doc comment.

https://github.com/adobe/aepsdk-core-ios/pull/299/files#diff-02bd1581c751b39213c003d305212478496f8a5290a1a877cef358a338c892f4R187

PR #585, opened 2021-03-24
Fetch latest shared state for nil event or an event that is yet to be versioned

nporter-adbe:
## Description

Based on discussions we want to get the latest shared state when a nil event is passed when fetching the shared state.

## Types of changes

## Checklist:


AEPCore/Sources/eventhub/EventHub.swift
```diff
@@ -214,7 +214,7 @@ final class EventHub {
     /// Retrieves the `SharedState` for a specific extension
     /// - Parameters:
     ///   - extensionName: An extension name whose `SharedState` will be returned
-    ///   - event: If not nil, will retrieve the `SharedState` that corresponds with this event's version. If nil will return the earliest `SharedState`
+    ///   - event: If not nil, will retrieve the `SharedState` that corresponds with this event's version or latest if not yet versioned. If event is nil will return the earliest `SharedState`
```
emdobrin, line 217: In Android and Objc core we return latest shared state when event is nil. I think we should make them consistent in that situation - this may be considered as a bug fix as this functionality existed before.

AEPCore/Tests/EventHubTests/EventHubTests.swift
```diff
@@ -922,6 +922,21 @@ class EventHubTests: XCTestCase {
...
+
+        // test
+        eventHub.createSharedState(extensionName: EventHubTests.MOCK_EXTENSION_NAME, data: SharedStateTestHelper.ONE, event: nil)
+        let event1 = Event(name: "test1", type: EventType.analytics, source: EventSource.requestContent, data: nil)
+        eventHub.dispatch(event: event1)
+        eventHub.createSharedState(extensionName: EventHubTests.MOCK_EXTENSION_NAME, data: SharedStateTestHelper.TWO, event: event1)
+
+        // verify
+        let unversionedEvent = Event(name: "test2", type: EventType.custom, source: EventSource.none, data: nil)
+        validateSharedState(EventHubTests.MOCK_EXTENSION_NAME, unversionedEvent, "two")
```
emdobrin, line 937: let's try again for the same event

eventHub.createSharedState(extensionName: EventHubTests.MOCK_EXTENSION_NAME, data: SharedStateTestHelper.THREE, event: event1)
validateSharedState(EventHubTests.MOCK_EXTENSION_NAME, unversionedEvent, "three")

AEPCore/Sources/eventhub/EventHub.swift
nporter-adbe: It's intended to return the earliest, it's the best way around the fact we don't have a boot even to fetch the latest. 
https://github.com/adobe/aepsdk-core-ios/issues/563

emdobrin: But this breaks this contract on the public getSharedState API https://github.com/adobe/aepsdk-core-ios/blob/main/AEPCore/Sources/eventhub/ExtensionRuntime.swift#L65

shalehaha: @emdobrin good catch, we should update the comment in the runtime class

emdobrin: I am concerned that this will cause issues for developers implementing both iOS and Android extensions, as the behavior is not consistent. The intention with null event for getSharedState was originally this, to always get latest, not earliest.

swarna04: Thanks for bringing this up @emdobrin. I had this same question when I was reviewing the Target swift code where we fetch and cache the latest available config.

nporter-adbe:
@shalehaha @emdobrin I'd like to get this in before the 3.1.0 release. Let me know if there are any changes needed.

shalehaha (approved):

emdobrin (approved):

PR #593, opened 2021-03-30
[AMSDK-11201] - Differentiate XDM vs standard shared state via eventName

PravinPK:
Different eventName to differentiate between XDM vs standard shared state change event

Standard state change event name = "Shared state change"
XDM state change event name = "Shared state change (XDM)"

Unit tests are updated

### DIFF
--- a/AEPCore/Sources/eventhub/EventHub.swift
+++ b/AEPCore/Sources/eventhub/EventHub.swift
@@ -180,7 +180,7 @@ final class EventHub {
             }
 
             sharedState.set(version: version, data: data)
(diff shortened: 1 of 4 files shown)


AEPCore/Sources/eventhub/EventHub.swift
```diff
@@ -180,7 +180,7 @@ final class EventHub {
             }
 
             sharedState.set(version: version, data: data)
-            self.dispatch(event: self.createSharedStateEvent(extensionName: extensionName))
+            self.dispatch(event: self.createSharedStateEvent(extensionName: extensionName, sharedStatetype: sharedStateType))
```
shalehaha, line 183: good catch!

Issue #596, opened 2021-04-05
Multiple EventHub shared states posted on registerExtensions(..)


emdobrin:
### Expected Behaviour
When calling MobileCore.registerExtension(extensions, completion) the MobileCore registers all the provided extensions, creates the Event Hub Shared state with version 0 with the registered extensions and starts the Event Hub for processing.

### Actual Behaviour
When calling MobileCore.registerExtension(extensions, completion) and using the compat layer the MobileCore posts multiple Event Hub shared states (see version 0 and 9 in the logs below).

This issue has the side effect of breaking the check for registered extensions from other extensions, such as Edge checking for the presence of Consent on boot. 

#### Steps to Reproduce
Run the Swift sample app and inspect the logs https://github.com/adobe/aepsdk-sample-app-ios/tree/main/Swift/AEPSampleApp

#### Platform and Version
See app's podfile 

#### Sample Code that illustrates the problem
See AppDelegate

#### Logs taken while reproducing problem

<details>
  <summary>Logs, click to expand</summary>

 2021-04-05 12:21:40.719347-0700 AEPSampleApp[71083:1263855] [AEP SDK DEBUG - <EventHub>] Shared state created for com.adobe.module.eventhub with version 0 and data: 
{
  "version" : "3.1.0",
  "extensions" : {
    "com.adobe.assurance" : {
      "version" : "1.1.0",
      "friendlyName" : "com.adobe.assurance"
    },
    "com.adobe.module.lifecycle" : {
      "friendlyName" : "Lifecycle",
      "version" : "3.1.0"
    }
  }
}
2021-04-05 12:21:40.733977-0700 AEPSampleApp[71083:1263870] [AEP SDK DEBUG - <EventHub>] Shared state created for com.adobe.module.eventhub with version 9 and data: 
{
  "version" : "3.1.0",
  "extensions" : {
    "com.adobe.edge.identity" : {
      "version" : "1.0.0",
      "friendlyName" : "Edge Identity"
    },
    "com.sampleCompany.module.sample" : {
      "version" : "1.0.0",
      "friendlyName" : "Sample"
    },
    "com.adobe.module.configuration" : {
      "version" : "3.1.0",
      "friendlyName" : "Configuration"
    },
    "com.adobe.module.signal" : {
      "friendlyName" : "Signal",
      "version" : "3.1.0"
    },
    "com.adobe.assurance" : {
      "friendlyName" : "com.adobe.assurance",
      "version" : "1.1.0"
[...]

PR #598, opened 2021-04-06
Only share event hub state after started

nporter-adbe:
## Description

If an extension is registered by itself via `registerExtension` before `registerExtensions` it causes an extra EventHub shared state to be dispatched. This PR adds an optimization to only share the EventHub shared state after it has been started.

I ran the sample app that was described in https://github.com/adobe/aepsdk-core-ios/issues/596 and confirmed that now only 1 EventHub shared state is published.

## Types of changes

## Checklist:

### DIFF
--- a/AEPCore/Sources/eventhub/EventHub.swift
+++ b/AEPCore/Sources/eventhub/EventHub.swift
@@ -29,6 +29,7 @@ final class EventHub {
     private var eventNumberCounter = AtomicCounter()
     private let eventQueue = OperationOrderer<Event>("EventHub")
     private var preprocessors = ThreadSafeArray<EventPreprocessor>(identifier: "com.adobe.eventHub.preprocessors.queue")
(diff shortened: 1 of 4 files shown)


AEPCore/Sources/eventhub/EventHub.swift
```diff
@@ -29,6 +29,7 @@ final class EventHub {
     private var eventNumberCounter = AtomicCounter()
     private let eventQueue = OperationOrderer<Event>("EventHub")
     private var preprocessors = ThreadSafeArray<EventPreprocessor>(identifier: "com.adobe.eventHub.preprocessors.queue")
+    private var started = false // true if the `EventHub` is started, false otherwise. Should only be accessed from within the `eventHubQueue`
```
nporter-adbe, line 32: @shalehaha This is a bit lightweight... should I instead introduce an `AtomicBool` or would that be overkill?

shalehaha: seems good to me

shalehaha (approved):

PR #620, opened 2021-04-29
Security improvement

yangyansong-adbe:


nporter-adbe:
@yangyansong-adbe should we make these changes in other repos?

yangyansong-adbe:
@nporter-adbe  yes, we'd better update it in other repos as well. I will shared it in the slack channel after we have released core builds.

PR #625, opened 2021-05-07
[AMSDK-11237] Identity: Handle reset identities request

nporter-adbe:
## Description

This PR adds support for the Identity extension to listen for and respond to the reset identities event. When the Identity extension receives this event several operations take place:

- Existing ECID and all other ids are cleared
- A force sync is kicked off which will generate a new ECID, send the new ECID and cleared ids to the server, and save these changes to persistence
- Creates a new Identity shared state with the updated ids
- Dispatches a reset complete event

## Types of changes

## Checklist:


AEPIdentity/Sources/IdentityState.swift
```diff
@@ -283,6 +283,30 @@ class IdentityState {
...
+    /// Saves identities to persistence and creates a new shared state and dispatches a new` resetComplete` event after operation completes.
+    /// - Parameters:
+    ///   - event: event which triggered the reset call
+    ///   - createSharedState: function which creates new shared states
+    ///   - eventDispatcher: function which dispatches a new `Event`
+    func resetIdentifiers(event: Event,
+                          createSharedState: ([String: Any], Event) -> Void,
+                          eventDispatcher: (Event) -> Void) {
+        // clear the properties
+        identityProperties = IdentityProperties()
```
kevinlind, line 296: Actually, this will reset the stored privacy status to unknown. Maybe we should check if privacy is optedout and if so, simply abort?

AEPIdentity/Sources/IdentityState.swift
```diff
@@ -283,6 +283,30 @@ class IdentityState {
...
+        // do a force sync to generate ECID, then save the properties to persistence.
+        _ = syncIdentifiers(event: event)
+        createSharedState(identityProperties.toEventData(), event)
+
+        // dispatch reset complete event
+        let event = Event(name: IdentityConstants.EventNames.RESET_IDENTITIES_COMPLETE,
+                          type: EventType.identity,
+                          source: EventSource.resetComplete,
+                          data: nil)
+        eventDispatcher(event)
```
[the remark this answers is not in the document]
nporter-adbe: Yeah, that's a fair point, I'll remove it and we can add it later if we find a use case.

AEPIdentity/Sources/IdentityState.swift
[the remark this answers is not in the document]
nporter-adbe: @emdobrin any thoughts on this?

AEPIdentity/Tests/IdentityStateTests.swift
```diff
@@ -987,6 +987,58 @@ class IdentityStateTests: XCTestCase {
...
+        // setup
+        let sharedStateExpectation = XCTestExpectation(description: "Shared state should be updated once")
+        sharedStateExpectation.isInverted = true
+        let configSharedState = [IdentityConstants.Configuration.EXPERIENCE_CLOUD_ORGID: "test-org",
+                                 IdentityConstants.Configuration.EXPERIENCE_CLOUD_SERVER: "test-server",
+                                 IdentityConstants.Configuration.GLOBAL_CONFIG_PRIVACY: PrivacyStatus.optedOut.rawValue] as [String: Any]
+        state.lastValidConfig = configSharedState
+        let startingEcid = ECID()
+        state.identityProperties.ecid = startingEcid
+        state.identityProperties.advertisingIdentifier = "test-ad-id"
```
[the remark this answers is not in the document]
nporter-adbe: I removed the portion where we set the identifiers, but it's still valid in the context of not queuing a hit.

AEPIdentity/Tests/IdentityFunctionalTests.swift
```diff
@@ -269,4 +269,28 @@ class IdentityFunctionalTests: XCTestCase {
...
+
+        let resetEvent = Event(name: "test reset event", type: EventType.genericIdentity, source: EventSource.requestReset, data: nil)
+
+        // test
+        mockRuntime.simulateComingEvent(event: resetEvent)
+
+        // verify
+        XCTAssertNotEqual(previousEcid.ecidString, identity.state?.identityProperties.ecid?.ecidString)
+        let sharedState = mockRuntime.createdSharedStates.first!
+        XCTAssertEqual(identity.state?.identityProperties.ecid?.ecidString, sharedState?[IdentityConstants.EventDataKeys.VISITOR_ID_ECID] as? String)
```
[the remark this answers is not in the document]
nporter-adbe: Are you suggesting that we have some way to ignore network responses for hits that have already been queued once a reset is received?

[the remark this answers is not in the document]
nporter-adbe: I added a safe-check here to only update the stored properties if the ECID in the response matches what we have stored locally. Writing the test you described with the network delay isn't really possible.
https://github.com/adobe/aepsdk-core-ios/pull/625/commits/3af00971db5b399e616cedd3fae80e9394cc1de0

AEPIdentity/Sources/IdentityState.swift
```diff
@@ -283,6 +283,30 @@ class IdentityState {
...
+        guard identityProperties.privacyStatus != .optedOut else { return }
+        // clear the properties
+        identityProperties.ecid = nil
+        identityProperties.advertisingIdentifier = nil
+        identityProperties.blob = nil
+        identityProperties.locationHint = nil
+        identityProperties.customerIds?.removeAll()
+        identityProperties.isAidSynced = false
+        identityProperties.pushIdentifier = nil
+        pushIdManager.updatePushId(pushId: nil)
```
emdobrin, line 302: This will generate a push optout hit which doesn't seem to be required since the old ECID will no longer be used for interactions, did you think otherwise when you included it?

nporter-adbe: DIdn't mean to include it. removed.

kevinlind: Calling updatePushId will dispatch a push enabled/disabled event to Analytics. If you don't call `updatePushId`, the `pushEnabled` persisted value isn't updated to false. It appears, if I'm reading it correctly, if push was enabled and we simply clear the push ID without updating the push manager, new push enabled events won't get sent to Analytics as `pushEnabled` is already true.
If this logic is needed when privacy is set to opt-out, where the old ECID is also no longer used, I don't see why we wouldn't do it for reset identities.

AEPIdentity/Sources/IdentityState.swift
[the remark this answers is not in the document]
kevinlind: The opt-out hit is sent after an ECID is cleared, so saying it doesn't need to be sent because there are no more interactions with the old ECID doesn't seem valid to me. I think the point of the opt-out hit is to delink the old ECID from the ad id server-side. However, I didn't consider the case of any existing hits which may be queued, so sending the opt-out hit before the queued hits are processed would be incorrect.
As we are clearing all the identifiers, are any existing queued hits still valid? In the case of privacy opt-out we took the position that those hits were no longer needed as we were clearing the identifiers anyway. Should we consider this case the same, and clear the hits?

AEPIdentity/Tests/IdentityFunctionalTests.swift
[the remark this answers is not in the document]
nporter-adbe: It's possible, but overkill for this.

AEPIdentity/Sources/IdentityState.swift
nporter-adbe: @kevinlind The original reason I had it was because we had that same logic for the opt-out so I assumed it applied as the two processes seem very similar. @emdobrin how does handle identities differentiate from the opt-out case here?

AEPIdentity/Sources/IdentityState.swift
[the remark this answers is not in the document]
nporter-adbe: @kevinlind I'd be open to also clearing the hits in this case.

AEPIntegrationTests/IdentityIntegrationTests.swift
```diff
@@ -204,4 +204,43 @@ class IdentityIntegrationTests: XCTestCase {
...
+        // set first ecid
+        var props = IdentityProperties()
+        let firstEcid = ECID()
+        props.ecid = firstEcid
+        props.saveToPersistence()
+
+        initExtensionsAndWait()
+
+        let requestExpectation = XCTestExpectation(description: "identity hits go out")
+        requestExpectation.expectedFulfillmentCount = 2
```
emdobrin, line 218: for the test expectation this should be 1, right? let's wait and clear any other network requests sent out before `MobileCore.resetIdentities()`

AEPIdentity/Tests/IdentityFunctionalTests.swift
```diff
@@ -269,4 +269,28 @@ class IdentityFunctionalTests: XCTestCase {
...
+        let previousEcid = ECID()
+        identity.state?.identityProperties.ecid = previousEcid
+
+        let resetEvent = Event(name: "test reset event", type: EventType.genericIdentity, source: EventSource.requestReset, data: nil)
+
+        // test
+        mockRuntime.simulateComingEvent(event: resetEvent)
+
+        // verify
+        XCTAssertNotEqual(previousEcid.ecidString, identity.state?.identityProperties.ecid?.ecidString)
```
kevinlind, line 292: Verify the new ECID is also not nil.

AEPIdentity/Tests/IdentityFunctionalTests.swift
```diff
@@ -269,4 +269,28 @@ class IdentityFunctionalTests: XCTestCase {
...
+        let sharedState = mockRuntime.createdSharedStates.first!
```
kevinlind, line 293: Verify the shared state only contains ECID (and last sync, as that's not reset).

AEPIdentity/Tests/IdentityStateTests.swift
```diff
@@ -987,6 +1027,56 @@ class IdentityStateTests: XCTestCase {
...
+
+        let resetEvent = Event(name: "test reset event", type: EventType.genericIdentity, source: EventSource.requestReset, data: nil)
+
+        // test
+        state.resetIdentifiers(event: resetEvent, createSharedState: { (data, _) in
+            sharedStateExpectation.fulfill()
+        })
+
+        // verify
+        XCTAssertTrue(mockHitQueue.queuedHits.isEmpty) // hit should NOT be queued in the hit queue
```
emdobrin, line 1077: let's also add some asserts for 
```
XCTAssertFalse(mockHitQueue.calledClear)
XCTAssertFalse(mockPushIdManager.calledResetPersistedFlags)
```

AEPIdentity/Tests/IdentityStateTests.swift
```diff
@@ -987,6 +1027,56 @@ class IdentityStateTests: XCTestCase {
...
+
+        // verify
+        XCTAssertTrue(mockHitQueue.calledClear)
+        XCTAssertTrue(mockPushIdManager.calledResetPersistedFlags)
+        wait(for: [sharedStateExpectation], timeout: 0.5)
+    }
+
+    func testResetIdentitiesOptedOut() {
+        // setup
+        let sharedStateExpectation = XCTestExpectation(description: "Shared state should be updated once")
```
emdobrin, line 1062: description -> "Shared state should not be updated"

AEPIdentity/Tests/IdentityStateTests.swift
```diff
@@ -987,6 +1027,56 @@ class IdentityStateTests: XCTestCase {
...
+
+        let configSharedState = [IdentityConstants.Configuration.EXPERIENCE_CLOUD_ORGID: "test-org",
+                                 IdentityConstants.Configuration.EXPERIENCE_CLOUD_SERVER: "test-server",
+                                 IdentityConstants.Configuration.GLOBAL_CONFIG_PRIVACY: PrivacyStatus.optedIn.rawValue] as [String: Any]
+        state.lastValidConfig = configSharedState
+        let startingEcid = ECID()
+        state.identityProperties.ecid = startingEcid
+        state.identityProperties.advertisingIdentifier = "test-ad-id"
```
emdobrin, line 1041: let's mock the values and assert the reset for all properties being updated: locationHint, blob, customerIds etc.

AEPIdentity/Tests/IdentityStateTests.swift
```diff
@@ -632,6 +633,42 @@ class IdentityStateTests: XCTestCase {
         XCTAssertEqual(hitResponse.ttl, state.identityProperties.ttl) // ttl should have been updated
     }
 
+    /// Tests that when a non-opt out response is handled with a non-matching ECID that we don't update the last sync and other identity properties, along with dispatching two identity events
```
emdobrin, line 636: let's add a comment here like:
` This situation can usually happen if a network response is handled at the same time as the resetIdentities request.`

emdobrin (commented):
@nporter-adbe looks good with a few comments on testing

PR #628, opened 2021-05-17
[MOB-14375] Add getDeviceModelNumber to system info service

nporter-adbe:
## Description

Adds a new API to the `SystemInfoService` to retrieve the model number for the current device. This is to support the model number field on the Device XDM schema: https://github.com/adobe/xdm/blob/master/docs/reference/datatypes/device.schema.md#device-properties

The output of this API is something like "D431AP", https://www.theiphonewiki.com/wiki/D431AP.

It is similar to getDeviceBuildId in Android, however, I found getDeviceBuildId wasn't a very descriptive API name so I opted for getDeviceModelNumber.

## How Has This Been Tested?

I tested this manually on my own device as this is a difficult function to write unit tests for.

## Types of changes

## Checklist:

### DIFF
--- a/AEPServices/Mocks/MockSystemInfoService.swift
+++ b/AEPServices/Mocks/MockSystemInfoService.swift
@@ -110,7 +110,7 @@ public class MockSystemInfoService: SystemInfoService {
     public func getDisplayInformation() -> (width: Int, height: Int) {
         return displayInformation
     }
-    
+
     public var deviceModelNumber: String = ""
     public func getDeviceModelNumber() -> String {
         return deviceModelNumber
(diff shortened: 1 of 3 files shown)


emdobrin:
@nporter-adbe is this api supported for tvos, macos or is this something we need to keep in mind and test when adding support for these platforms?
On release, let's remember to bump the minor version for AEPServices for this new API.

nporter-adbe:
@emdobrin I know this is supported in MacOS, but I'd have to check for tvOS as I don't have an Apple TV to confirm. Regardless there are a few considerations we'll need to make regarding the SystemInfoService when adding support for these platforms in the future and I think this and the usage of hw.machine will be some of them.

PR #629, opened 2021-05-18
[MOB-14325] - Adding few more essential API to ThreadSafePrimitives

PravinPK:
ThreadSafeArray
 - API to remove the first element

ThreadSafeDictionary
 - API to retrieve all keys

### DIFF
--- a/AEPServices/Sources/utility/ThreadSafeArray.swift
+++ b/AEPServices/Sources/utility/ThreadSafeArray.swift
@@ -35,6 +35,17 @@ public final class ThreadSafeArray<T> {
             self.array.removeAll()
         }
     }
+    
+    /// Removes and returns the first element of the thread safe array.
+    /// Return nil if the array is empty
+    public func removeFirst() -> T? {
+        queue.sync {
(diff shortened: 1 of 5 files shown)


AEPServices/Sources/utility/ThreadSafeArray.swift
```diff
@@ -36,6 +36,17 @@ public final class ThreadSafeArray<T> {
...
     }
 
+    /// Removes and returns the first element of the thread safe array.
+    /// Returns nil if the array is empty
+    public func removeFirst() -> T? {
+        queue.sync {
+            if array.isEmpty {
+                return nil
+            }
+            return self.array.removeFirst()
```
shalehaha, line 46: what's the behavior of this method if `self.array` is empty?

PravinPK: It crashes :(

PravinPK: And I realized better to handle the crash and return optional

PR #631, opened 2021-05-19
[MOB-14383] - Update NetworkService to post raw blob data

PravinPK:
**Premises for the change** : 
Ability to create/make post network request with raw data (data that is not UTF encoded). Usage is for Assurance to upload screenshots to blob service and many more

**Problem with current API**
The current API takes the `connectionPayload` as a String converts to utf8 formatted Data. Which is not suitable for sending images 

**Code changes**
- `Connectionpayload` property of `NetworkRequest` class is now stored as Data as opposed to String
- New initializer API for `NetworkRequest`

**Things considered**
- Do not change the existing API that coz of backwards compatibility.
-  Defaulting the API that takes `connectPayloadData` to use HTTP post method. (I am on wall with this one, let me know)

### DIFF
--- a/AEPServices/Sources/network/NetworkRequest.swift
+++ b/AEPServices/Sources/network/NetworkRequest.swift
@@ -33,30 +33,24 @@ import Foundation
     ///   - connectTimeout: optional connect timeout value in seconds; default is 5 seconds
     ///   - readTimeout: optional read timeout value in seconds, used to wait for a read to finish after a successful connect, default is 5 seconds
     /// - Returns: an initialized `NetworkRequest` object
-    public init(url: URL, httpMethod: HttpMethod = HttpMethod.get, connectPayload: String = "", httpHeaders: [String: String] = [:], connectTimeout: TimeInterval = 5, readTimeout: TimeInterval = 5) {
-        self.url = url
-        self.httpMethod = httpMethod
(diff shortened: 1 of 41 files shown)


AEPServices/Sources/network/NetworkRequest.swift
```diff
@@ -36,7 +36,30 @@ import Foundation
...
+    /// Initialize the `NetworkRequest` to post data.
+    /// The instance created from this initializer default to making http post network call.
+    /// - Parameters:
+    ///   - url: URL used to initiate the network connection, should use https scheme
+    ///   - connectPayloadData: the body of the network request as a String; this parameter is ignored for GET requests
+    ///   - httpHeaders: optional HTTP headers for the request
+    ///   - connectTimeout: optional connect timeout value in seconds; default is 5 seconds
+    ///   - readTimeout: optional read timeout value in seconds, used to wait for a read to finish after a successful connect, default is 5 seconds
+    /// - Returns: an initialized `NetworkRequest` object
+    public init(url: URL, connectPayloadData: Data = Data(), httpHeaders: [String: String] = [:], connectTimeout: TimeInterval = 5, readTimeout: TimeInterval = 5) {
```
nporter-adbe, line 59: This looks good, my only comment is we should make a helper function to perform common init steps. As there is quite a bit of overlap between both `init` functions.

nporter-adbe (approved):

PR #633, opened 2021-05-20
[MOB-14384] - API to modify Floating button image

PravinPK:


AEPServices/Sources/ui/floating/FloatingButton.swift
```diff
@@ -63,6 +63,17 @@ public class FloatingButton: NSObject, FloatingButtonPresentable {
         }
     }
 
+    /// Set the Image for the floating button.
+    /// The size of the floating button is 60x60 (width x height), provide the image data accordingly
+    /// - Parameters:
+    ///     - imageData : The `Data` representation of a UIImage
+    public func setButtonImage(imageData: Data) {
+        DispatchQueue.main.async {
+            let image = UIImage(data: imageData)
```
nporter-adbe, line 72: Any reason we can't just take in a `UIImage` instead of `Data`? If not, I wonder if creating the `UIImage` off the main thread may be slightly more responsible.

AEPServices/Sources/ui/floating/FloatingButton.swift
```diff
@@ -63,6 +63,17 @@ public class FloatingButton: NSObject, FloatingButtonPresentable {
         }
     }
 
+    /// Set the Image for the floating button.
```
nporter-adbe, line 66: My personal opinion here: But I don't think it is necessary to include doc comments on functions where the doc comment is defined in the protocol for the function.

PravinPK: Agreed it was repetitive

AEPServices/Sources/ui/floating/FloatingButton.swift
PravinPK: taking in `Data` is more generic than UIImage. Given that UIImage is available in TVOS and iOS and not in watchOS (WKInterfaceImage)... Let me know otherwise (if there is any benefits passing UIImage)

+ 1 for moving UIImage creation outside the main thread

nporter-adbe (approved):

PR #637, opened 2021-05-26
[MOB-14414] - API to set the starting position of floating button

PravinPK:
API to set the initial position of the floating button

**Usage**
```
        let floatingButton = ServiceProvider.shared.uiService.createFloatingButton(listener: self)
        floatingButton.setInitialPosition(position: FloatingButtonPosition.topRight)
        floatingButton.show()
```

**How it appears**

Screen Shot 2021-05-26 at 11 51 35 AM

**Supported Initial position values**
- center (default)
- topRight
- topLeft

### DIFF
--- a/AEPServices/Sources/ui/floating/FloatingButton.swift
+++ b/AEPServices/Sources/ui/floating/FloatingButton.swift
@@ -19,13 +19,15 @@ public class FloatingButton: NSObject, FloatingButtonPresentable {
 
     private let LOG_PREFIX = "FloatingButton"
 
-    private let PREVIEW_BUTTON_WIDTH = 60
-    private let PREVIEW_BUTTON_HEIGHT = 60
+    internal static let BUTTON_TOP_MARGIN = 40
+    internal static let PREVIEW_BUTTON_WIDTH = 60
+    internal static let PREVIEW_BUTTON_HEIGHT = 60
 
     private var singleTap: UITapGestureRecognizer?
     private var panner: UIPanGestureRecognizer?
     private var timer: Timer?
     private var buttonImageView: UIImageView?
+    private var buttonPosition: FloatingButtonPosition = .center
 
     private var listener: FloatingButtonDelegate?
 
@@ -68,6 +70,10 @@ public class FloatingButton: NSObject, FloatingButtonPresentable {
         DispatchQueue.main.async {
             self.buttonImageView?.image = image
         }
+    }
+
+    public func setInitialPosition(position: FloatingButtonPosition) {
+        buttonPosition = position
     }
 
     private func initFloatingButton() -> Bool {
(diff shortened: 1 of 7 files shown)


AEPServices/Sources/ui/floating/FloatingButtonPosition.swift
```diff
@@ -0,0 +1,44 @@
...
+ Unless required by applicable law or agreed to in writing, software distributed under
+ the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ OF ANY KIND, either express or implied. See the License for the specific language
+ governing permissions and limitations under the License.
+ */
+
+import Foundation
+import UIKit
+
+@objc(AEPFloatingButtonPosition) public enum FloatingButtonPosition: Int {
```
ravjain-adb, line 16: This will be exposed to customers also do we want to do that?

PravinPK: Yes this was intended, for anyone who creates a floatingButton to change its initial show position

PravinPK:
Another option to expose this functionality is to create another API in `UIServices` protocol

```
func createFloatingButton(listener: FloatingButtonDelegate, withInitialPosition: FloatingButtonPosition) -> FloatingButtonPresentable
```

Making this change by including the protocol method will have all the extensions mocking UIServices to implement this method

AEPServices/Sources/ui/floating/FloatingButtonPresentable.swift
```diff
@@ -23,4 +23,11 @@ public protocol FloatingButtonPresentable: Showable, Dismissible {
     /// - Parameters:
     ///     - imageData : The `Data` representation of a UIImage
     func setButtonImage(imageData: Data)
+
+    /// Set the initial position of floating button.
+    /// By default the initial position is set to `FloatingButtonPosition.center`.
+    /// Call this method before calling `floatingButton.show()` to set the position of the floating button when it appears.
+    /// - Parameters:
+    ///     - position : The `FloatingButtonPosition` defining the initial position of the floating button.
+    func setInitialPosition(position: FloatingButtonPosition)
```
nporter-adbe, line 32: Nit: I think we should name this API `setInitialPosition(position` to `setInitial(position` to allow the argument label to finish the name of the API. This is somewhat of a standard in Swift.

So the call would be like `setInitial(position: ...)` instead of `setInitialPosition(position: ...)`. Just slightly less repeteative.

AEPServices/Sources/ui/floating/FloatingButton.swift
```diff
@@ -113,14 +119,12 @@ public class FloatingButton: NSObject, FloatingButtonPresentable {
...
-                Log.debug(label: LOG_PREFIX, "Floating button couldn't be displayed, screenbound is nil.")
-                return false
-            }
-            UIView.animate(withDuration: 0.3, animations: {
-                var finalFrame = buttonImageView.frame
-                finalFrame.origin.x = (screenBounds.width - CGFloat(self.PREVIEW_BUTTON_WIDTH)) / 2
-                buttonImageView.frame = finalFrame
+
+            // set the initial position for animation
+            buttonImageView.frame.origin.x = buttonImageView.frame.origin.x + CGFloat(FloatingButton.PREVIEW_BUTTON_WIDTH)
```
nporter-adbe, line 124: Can we just use `+=` here?

AEPServices/Sources/ui/floating/FloatingButton.swift
```diff
@@ -113,14 +119,12 @@ public class FloatingButton: NSObject, FloatingButtonPresentable {
...
+            // animate the x-axis of the button to its original position
+            UIView.animate(withDuration: 0.3, delay: 0, options: .curveEaseIn, animations: {
+                buttonImageView.frame.origin.x = buttonImageView.frame.origin.x - CGFloat(FloatingButton.PREVIEW_BUTTON_WIDTH)
```
nporter-adbe, line 127: Same with `-=` here?

nporter-adbe (approved):

PR #640, opened 2021-05-28
[MOB-14426] - API to hide fullscreen message

PravinPK:
The motivation for this pull request is from Assurance, which wanted the fullScreen message to be taken off the screen but should still be in memory so it can be reshown from its previous state.

Example: 
The Status Screen for Assurance is now a FullScreenMessage. And this UI appears each time the user clicks on the Assurance floating button. In order to preserve the clientLogs that are populated in HTML. We were needed to implement the hide feature for the InAppMessage

Screen Shot 2021-05-28 at 3 09 03 PM

Note : The hide() API doesn't conform to any protocol. And can only be used by user who holds the FullScreenMessage object

### DIFF
--- a/AEPServices/Sources/ui/fullscreen/FullscreenMessage.swift
+++ b/AEPServices/Sources/ui/fullscreen/FullscreenMessage.swift
@@ -50,7 +50,7 @@ public class FullscreenMessage: NSObject, WKNavigationDelegate, FullscreenPresen
         self.isLocalImageUsed = isLocalImageUsed
         self.messageMonitor = messageMonitor
     }
-    
+
     /// Call this API to hide the fullscreen message.
     /// This API hides the fullscreen message with an animation, but it keeps alive its webView for future reappearances.
     /// Invoking show on a hidden fullscreen message, will display the fullscreen message in the existing state (i.e webView is not re-rendered)
--- a/AEPServices/Sources/ui/fullscreen/FullscreenMessage.swift
+++ b/AEPServices/Sources/ui/fullscreen/FullscreenMessage.swift
@@ -50,7 +50,12 @@ public class FullscreenMessage: NSObject, WKNavigationDelegate, FullscreenPresen
         self.isLocalImageUsed = isLocalImageUsed
         self.messageMonitor = messageMonitor
     }
-
+    
+    /// Call this API to hide the fullscreen message.
+    /// This API hides the fullscreen message with an animation, but it keeps alive its webView for future reappearances.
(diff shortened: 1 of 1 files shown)


AEPServices/Sources/ui/fullscreen/FullscreenMessage.swift
```diff
@@ -51,12 +51,27 @@ public class FullscreenMessage: NSObject, WKNavigationDelegate, FullscreenPresen
...
 
+    public func hide() {
+        DispatchQueue.main.async {
+            if self.messageMonitor.dismiss() ==  false {
+                return
+            }
+            self.dismissWithAnimation(animate: true, shouldDeallocateWebView: false)
+        }
+    }
+
```
PravinPK, line 62: In this function it is mandatory to call the `messageMonitor` to dismiss. So forth other InAppMessages can be shown when the current one is hidden

nporter-adbe: If this is not defined in a protocol can you add doc comments for it?

AEPServices/Sources/ui/fullscreen/FullscreenMessage.swift
```diff
@@ -195,7 +216,9 @@ public class FullscreenMessage: NSObject, WKNavigationDelegate, FullscreenPresen
                 self.webView?.frame = newFrame
             }) { _ in
                 self.webView?.removeFromSuperview()
-                self.webView = nil
+                if shouldDeallocateWebView {
```
nporter-adbe, line 219: Is this complexity worth it? Why not just create a new web view each time you need to show it?

PravinPK: From Assurance point of view, it was worth it. Especially with `AssuranceStatusUI` which is supposed to appear and reappear as the user taps the `AssuranceFloatingbutton`

It saved me from saving the state of the StatusUI
And also reusing webView minimized the number of javascript commands I have to run every time when it kicks off.. or to load the webview to the current state

Also thinking beyond, the `hide` feature opens up a flexible usage of FullScreenMessage... Example(rendering videos / filling up online form).. where getting rid of the fullscreen message temporarily without losing the current state will be a bonus

AEPServices/Sources/ui/fullscreen/FullscreenMessage.swift
PravinPK: yeap that should be done

nporter-adbe (approved):

PR #644, opened 2021-06-04
[MOB-14460] - Fullscreen message API webViewFinishLoading

PravinPK:
- Included an option protocol method FullScreenMessageDelegate
```
optional func webViewDidFinishLoading()
```

This method will get invoked when the webview has finished loading its original content.
This change is non-breaking (since the protocol method is optional).

**Testing**
1. Tested with Assurance to verify if the method is called aptly after web loading
2. Tested with TargetPreview to verify this optional method in protocol doesn't create compile error for existing FullScrennMessageDelegate Handlers.


AEPServices/Sources/ui/fullscreen/FullscreenMessageDelegate.swift
```diff
@@ -34,6 +34,10 @@ import Foundation
     @objc(overrideUrlLoadFullscreenMessage:url:)
     func overrideUrlLoad(message: FullscreenMessage, url: String?) -> Bool
 
+    /// Invoked when the fullscreen message finished loading the webView.
+    @objc(webViewDidFinishLoading)
+    optional func webViewDidFinishLoading()
```
nporter-adbe, line 39: In iOS we typically pass in the object that is being referenced, so I think I'd like to see something like:
`webViewDidFinishLoading(webView: WKWebView)` for this protocol definition.

For example, if you take a look at the `UITableViewDelegate` you will notice that the `UITableView` in reference is nearly always passed into the delegate function.
https://developer.apple.com/documentation/uikit/uitableviewdelegate

AEPServices/Sources/ui/fullscreen/FullscreenMessage.swift
```diff
@@ -30,6 +30,7 @@ public class FullscreenMessage: NSObject, WKNavigationDelegate, FullscreenPresen
     var listener: FullscreenMessageDelegate?
```
nporter-adbe, line 30: Not related to your changes but for memory sake, we need to make this `weak` to avoid a retain cycle. Let's also rename it to `delegate`.

`weak var delegate: FullscreenMessageDelegate?`

https://medium.com/macoclock/delegate-retain-cycle-in-swift-4d9c813d0544

AEPServices/Sources/ui/fullscreen/FullscreenMessage.swift
```diff
@@ -182,6 +183,15 @@ public class FullscreenMessage: NSObject, WKNavigationDelegate, FullscreenPresen
         }
     }
 
+    /// Delegate method invoked when the webView navigation is complete.
+    public func webView(_ webView: WKWebView, didFinish navigation: WKNavigation!) {
+        if navigation == self.loadingNavigation {
+            if let funcBlock = self.listener?.webViewDidFinishLoading {
```
nporter-adbe, line 189: You should be able to just do `self.delegate?.webViewDidFinishLoading(webView: webView)`

AEPServices/Sources/ui/fullscreen/FullscreenMessage.swift
```diff
@@ -30,6 +30,7 @@ public class FullscreenMessage: NSObject, WKNavigationDelegate, FullscreenPresen
     var listener: FullscreenMessageDelegate?
     public private(set) var webView: UIView?
     private var messageMonitor: MessageMonitoring
+    private var loadingNavigation: WKNavigation?
```
nporter-adbe, line 33: This is an interesting approach... typically I've seen it done like:

1. Add the WKNavigationDelegate protocol to your class declaration, like: `class FullscreenMessage: WKNavigationDelegate`
2. Set the `navigationDelegate` property of `webView` to `self,` before loading the web request, like: `webView?.navigationDelegate = self`
3. Then implement `webView(_:didFinish:)`

The approach you've laid out should be fine but isn't one I've seen used before.

nporter-adbe (changes requested):

AEPServices/Sources/ui/fullscreen/FullscreenMessage.swift
PravinPK: Isn't this current approach the same that you have mentioned ?
The only extra thing that's is happening is we filter out all the other page loads from `webView(_:didFinish:)`

setting the loadNavigation

nporter-adbe: Good point, I didn't notice the class conforms to `WKNavigationDelegate`.

AEPServices/Sources/ui/fullscreen/FullscreenMessage.swift
PravinPK: Yes this can be simplified since there is no logic for the else block (when optional protocol method is not available)

AEPServices/Sources/ui/fullscreen/FullscreenMessage.swift
PravinPK: Decided to stick with the `listener`  as opposed to  Apple friendly `delegate`... since  the publicAPI `CreateFullScreenMessage`   already calls this variable  `listener`... Switching only the internal variable name to  delegate might end up being confusing

nporter-adbe (approved):

PR #649, opened 2021-06-15
Fix authentication state when calling syncIdentifiers

kevinlind:
## Description

Send authentication state as MobileVisitorAuthenticationState instead of Int in event data for syncIdentifiers API call. The Identity extension casts the authentication state as a MobileVisitorAuthenticationState when extracting from the event data. When using the raw value (Int) of the MobileVisitorAuthenticationState enum, the cast always fails and returns value .unknown. This fix will allow the user specified authentication state for the custom identifier in the syncIdentifiers call instead of all authentication states being unknown.

Internal ticket: AMSDK-11493

## Types of changes

## Checklist:

### DIFF
--- a/AEPIdentity/Sources/Identity+PublicAPI.swift
+++ b/AEPIdentity/Sources/Identity+PublicAPI.swift
@@ -112,7 +112,7 @@ import Foundation
     static func syncIdentifiers(identifiers: [String: String]?, authenticationState: MobileVisitorAuthenticationState) {
         var eventData = [String: Any]()
         eventData[IdentityConstants.EventDataKeys.IDENTIFIERS] = identifiers
-        eventData[IdentityConstants.EventDataKeys.AUTHENTICATION_STATE] = authenticationState.rawValue
+        eventData[IdentityConstants.EventDataKeys.AUTHENTICATION_STATE] = authenticationState
         eventData[IdentityConstants.EventDataKeys.FORCE_SYNC] = false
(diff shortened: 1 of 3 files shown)


AEPIdentity/Sources/Identity+PublicAPI.swift
```diff
@@ -112,7 +112,7 @@ import Foundation
     static func syncIdentifiers(identifiers: [String: String]?, authenticationState: MobileVisitorAuthenticationState) {
         var eventData = [String: Any]()
         eventData[IdentityConstants.EventDataKeys.IDENTIFIERS] = identifiers
-        eventData[IdentityConstants.EventDataKeys.AUTHENTICATION_STATE] = authenticationState.rawValue
+        eventData[IdentityConstants.EventDataKeys.AUTHENTICATION_STATE] = authenticationState
```
kevinlind, line 115: Setting the `MobileVisitorAuthenticationState` in the event data causes the logged event to just display the enum name instead of its value:
```
data:  {
    authenticationstate = "AEPIdentity.MobileVisitorAuthenticationState";
    forcesync = 0;
    issyncevent = 1;
    visitoridentifiers =     {
        idType1 = 1234567;
    };
```
An alternative to this fix is to continue to set the raw value in the Event but change Event+Identity to cast the value as an Int and then convert to a MobileVisitorAuthenticationState when reading the event data.

nporter-adbe (approved):

PR #658, opened 2021-06-25
[AMSDK-11514] Prevent possible crash at shutdown in EventHub

nporter-adbe:
## Description

We received a crash report where it appears we are making a reference to the EventHub after it has been deallocated resulting in a crash. The most likely fix is to use `weak` when referring to `self` in the EventHub when in a closure.

https://jira.corp.adobe.com/browse/AMSDK-11514

## Types of changes

## Checklist:

### DIFF
--- a/AEPCore/Sources/eventhub/EventHub.swift
+++ b/AEPCore/Sources/eventhub/EventHub.swift
@@ -45,15 +45,12 @@ final class EventHub {
         registerExtension(EventHubPlaceholderExtension.self, completion: { _ in })
 
         // Setup eventQueue handler for the main OperationOrderer
-        eventQueue.setHandler { (event) -> Bool in
-
-            let processedEvent = self.preprocessors.shallowCopy.reduce(event) { event, preprocessor in
-                preprocessor(event)
-            }
+        eventQueue.setHandler { [weak self] (event) -> Bool in
+            guard let processedEvent = self?.preprocessors.shallowCopy.reduce(event, {$1($0)}) else { return true }
 
             // Handle response event listeners first
             if let responseID = processedEvent.responseID {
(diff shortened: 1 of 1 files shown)


AEPCore/Sources/eventhub/EventHub.swift
```diff
@@ -45,15 +45,12 @@ final class EventHub {
         registerExtension(EventHubPlaceholderExtension.self, completion: { _ in })
 
         // Setup eventQueue handler for the main OperationOrderer
-        eventQueue.setHandler { (event) -> Bool in
-
-            let processedEvent = self.preprocessors.shallowCopy.reduce(event) { event, preprocessor in
-                preprocessor(event)
-            }
+        eventQueue.setHandler { [weak self] (event) -> Bool in
+            guard let processedEvent = self?.preprocessors.shallowCopy.reduce(event, {$1($0)}) else { return true }
```
cdhoffmann, line 49: Rather than use an optional self (to avoid the nil-coalescing) I think unwrapping self would be best here

`guard let self = self else { return }`

cdhoffmann (approved):

AEPCore/Sources/eventhub/EventHub.swift
nporter-adbe: I think this suggestion is probably fine, but I am slightly concerned that with the suggestion, we're still just holding a strong reference to self which could result in the same issue we're seeing. I'd _slightly_ prefer to just weakly reference self.

Thoughts?

cdhoffmann: Hmmm. I guess I've never fully questioned the weak-strong dance which has become the standard in these use cases. Let me think about it a bit more. But I'm cool with this going in. 👍🏽

PR #672, opened 2021-07-28
[AMSDK-9271] Dispatch Lifecycle XDM events application launch & application close

addb:
## Types of changes

## Checklist:

### DIFF
--- a/AEPLifecycle/Sources/Lifecycle.swift
+++ b/AEPLifecycle/Sources/Lifecycle.swift
@@ -68,7 +68,7 @@ public class Lifecycle: NSObject, Extension {
             start(event: event, configurationSharedState: configurationSharedState)
         } else if event.isLifecyclePauseEvent {
             Log.debug(label: LifecycleConstants.LOG_TAG, "Pausing lifecycle.")
-            lifecycleState.pause(pauseDate: event.timestamp)
+            lifecycleState.pause(pauseDate: event.timestamp, dispatchApplicationCloseEvent: dispatchApplicationClose(xdm:))
         }
     }
 
@@ -82,7 +82,7 @@ public class Lifecycle: NSObject, Extension {
         let prevSessionInfo = lifecycleState.start(date: event.timestamp,
                                                    additionalContextData: event.additionalData,
                                                    adId: getAdvertisingIdentifier(event: event),
-                                                   sessionTimeout: getSessionTimeoutLength(configurationSharedState: configurationSharedState.value))
+                                                   sessionTimeout: getSessionTimeoutLength(configurationSharedState: configurationSharedState.value), dispatchApplicationLaunchEvent: dispatchApplicationLaunch(xdm:data:))
         updateSharedState(event: event, data: lifecycleState.getContextData()?.toEventData() ?? [:])
 
         if let prevSessionInfo = prevSessionInfo {
@@ -134,6 +134,32 @@ public class Lifecycle: NSObject, Extension {
                                data: eventData)
         Log.trace(label: LifecycleConstants.LOG_TAG, "Dispatching lifecycle start event with data: \n\(PrettyDictionary.prettify(eventData))")
         dispatch(event: startEvent)
+    }
+
+    /// Dispatches a Lifecycle application launch event with appropriate event data
+    /// - Parameters:
+    ///   - xdm: xdm data for the application launch event
+    ///   - data: current Lifecycle context data
+    private func dispatchApplicationLaunch(xdm: [String: Any], data: [String: Any]) {
+        let eventData: [String: Any] = [
+            LifecycleConstants.EventDataKeys.XDM: xdm,
+            LifecycleConstants.EventDataKeys.DATA: data
+        ]
+
+        let applicationLaunchEvent = Event(name: LifecycleConstants.EventNames.APPLICATION_LAUNCH, type: EventType.lifecycle, source: EventSource.applicationLaunch, data: eventData)
+        dispatch(event: applicationLaunchEvent)
+    }
+
+    /// Dispatches a Lifecycle application close event with appropriate event data
+    /// - Parameters:
+    ///   - xdm: xdm data for the application close event
+    private func dispatchApplicationClose(xdm: [String: Any]) {
+        let eventData: [String: Any] = [
+            LifecycleConstants.EventDataKeys.XDM: xdm
+        ]
+
+        let applicationCloseEvent = Event(name: LifecycleConstants.EventNames.APPLICATION_CLOSE, type: EventType.lifecycle, source: EventSource.applicationClose, data: eventData)
+        dispatch(event: applicationCloseEvent)
     }
 
     /// Reads the session timeout from the configuration shared state, if not found returns the default session timeout
--- a/AEPLifecycle/Sources/LifecycleState.swift
+++ b/AEPLifecycle/Sources/LifecycleState.swift
@@ -56,7 +56,7 @@ struct LifecycleState {
     ///   - adId: The advertising identifier provided by the identity extension
     ///   - sessionTimeout: The session timeout for this start event, defaults to 300 seconds
     /// - Returns: The previous session info if exists, otherwise nil
-    mutating func start(date: Date, additionalContextData: [String: Any]?, adId: String?, sessionTimeout: TimeInterval = TimeInterval(LifecycleConstants.DEFAULT_LIFECYCLE_TIMEOUT)) -> LifecycleSessionInfo? {
+    mutating func start(date: Date, additionalContextData: [String: Any]?, adId: String?, sessionTimeout: TimeInterval = TimeInterval(LifecycleConstants.DEFAULT_LIFECYCLE_TIMEOUT), dispatchApplicationLaunchEvent: @escaping ([String: Any], [String: Any]) -> Void) -> LifecycleSessionInfo? {
         let sessionContainer: LifecyclePersistedContext? = dataStore.getObject(key: LifecycleConstants.DataStoreKeys.PERSISTED_CONTEXT)
         // Build default LifecycleMetrics
         let metricsBuilder = LifecycleMetricsBuilder(dataStore: dataStore, date: date).addDeviceData()
@@ -95,6 +95,8 @@ struct LifecycleState {
             .addEnvironmentData()
             .addDeviceData()
 
+        dispatchApplicationLaunchEvent(xdmMetricsBuilder.buildXDMAppLaunchEventData() ?? [:], additionalContextData ?? [:])
+
         persistLifecycleContextData(startDate: date)
 
         return previousSessionInfo
@@ -102,7 +104,7 @@ struct LifecycleState {
 
     /// Pauses the current lifecycle session
     /// - Parameter pauseDate: date at which the pause event occurred
-    mutating func pause(pauseDate: Date) {
+    mutating func pause(pauseDate: Date, dispatchApplicationCloseEvent: @escaping ([String: Any]) -> Void ) {
         lifecycleSession.pause(pauseDate: pauseDate)
     }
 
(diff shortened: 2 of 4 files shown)


AEPLifecycle/Sources/Lifecycle.swift
```diff
@@ -82,7 +82,7 @@ public class Lifecycle: NSObject, Extension {
         let prevSessionInfo = lifecycleState.start(date: event.timestamp,
                                                    additionalContextData: event.additionalData,
                                                    adId: getAdvertisingIdentifier(event: event),
-                                                   sessionTimeout: getSessionTimeoutLength(configurationSharedState: configurationSharedState.value))
+                                                   sessionTimeout: getSessionTimeoutLength(configurationSharedState: configurationSharedState.value), dispatchApplicationLaunchEvent: dispatchApplicationLaunch(xdm:data:))
```
emdobrin, line 85: I would prefer not to escape the dispatchers to LifecycleState. I seems like we need to update LifecycleState to allow for fetching regular vs xdm content so we can dispatch the two events from here.
This way we can also update the order of the events, Lifecycle Response Content first, then Lifecycle Application Launch.

AEPLifecycle/Sources/Lifecycle.swift
```diff
@@ -136,6 +136,32 @@ public class Lifecycle: NSObject, Extension {
         dispatch(event: startEvent)
     }
 
+    /// Dispatches a Lifecycle application launch event with appropriate event data
+    /// - Parameters:
+    ///   - xdm: xdm data for the application launch event
+    ///   - data: current Lifecycle context data
```
emdobrin, line 142: data: additional free-form context data

AEPLifecycle/Sources/Lifecycle.swift
```diff
@@ -68,7 +68,7 @@ public class Lifecycle: NSObject, Extension {
             start(event: event, configurationSharedState: configurationSharedState)
         } else if event.isLifecyclePauseEvent {
             Log.debug(label: LifecycleConstants.LOG_TAG, "Pausing lifecycle.")
-            lifecycleState.pause(pauseDate: event.timestamp)
+            lifecycleState.pause(pauseDate: event.timestamp, dispatchApplicationCloseEvent: dispatchApplicationClose(xdm:))
```
emdobrin, line 71: Same comment here, let's not escape the dispatcher and call it from this class.

AEPLifecycle/Tests/FunctionalTests/LifecycleFunctionalTests.swift
```diff
@@ -192,16 +193,20 @@ class LifecycleFunctionalTests: XCTestCase {
         mockRuntime.simulateComingEvents(startEvent1, pauseEvent, startEvent2)
 
         // verify
-        XCTAssertEqual(2, mockRuntime.dispatchedEvents.count)
+        XCTAssertEqual(5, mockRuntime.dispatchedEvents.count)
```
emdobrin, line 196: Can you please add a comment with the expected events

AEPLifecycle/Tests/FunctionalTests/LifecycleFunctionalTests.swift
```diff
@@ -192,16 +193,20 @@ class LifecycleFunctionalTests: XCTestCase {
...
+
+
+        XCTAssertEqual("1", (dispatchedLifecycleStartEvent1.data?["lifecyclecontextdata"] as? [String: Any])?["launches"] as? String)
+        XCTAssertEqual("2", (dispatchedLifecycleStartEvent2.data?["lifecyclecontextdata"] as? [String: Any])?["launches"] as? String)
+
+        XCTAssertEqual(1_595_909_459, dispatchedLifecycleStartEvent2.data?["previoussessionstarttimestampmillis"] as? Double)
+        XCTAssertEqual(1_595_909_469, dispatchedLifecycleStartEvent2.data?["previoussessionpausetimestampmillis"] as? Double)
+        XCTAssertEqual(86400.0 * 7.0, dispatchedLifecycleStartEvent2.data?["maxsessionlength"] as? Double)
+        XCTAssertEqual(1_595_909_499, dispatchedLifecycleStartEvent2.data?["starttimestampmillis"] as? Double)
+        XCTAssertEqual("start", dispatchedLifecycleStartEvent2.data?["sessionevent"] as? String)
```
emdobrin, line 209: We should either update this test to cover all events or split into two tests, one covering the regular events and the other covering the XDM events.

AEPLifecycle/Tests/FunctionalTests/LifecycleFunctionalTests.swift
```diff
@@ -224,8 +229,8 @@ class LifecycleFunctionalTests: XCTestCase {
         mockRuntimeSession2.simulateComingEvents(createStartEvent())
 
         // verify
-        XCTAssertEqual(1, mockRuntimeSession2.dispatchedEvents.count)
-        XCTAssertEqual("CrashEvent", (mockRuntimeSession2.dispatchedEvents[0].data?["lifecyclecontextdata"] as? [String: Any])?["crashevent"] as? String)
+        XCTAssertEqual(2, mockRuntimeSession2.dispatchedEvents.count)
```
emdobrin, line 232: Can we instead ignore the XDM events for these tests?

emdobrin (commented):
@addb left a few comments below. Additionally can you please add new tests for the new events you added?

emdobrin:
@addb we should merge https://github.com/adobe/aepsdk-core-ios/pull/694 before this PR as it generates few changes.

AEPLifecycle/Sources/Lifecycle.swift
```diff
@@ -32,7 +32,7 @@ public class Lifecycle: NSObject, Extension {
     public required init(runtime: ExtensionRuntime) {
         self.runtime = runtime
         lifecycleState = LifecycleState(dataStore: NamedCollectionDataStore(name: name))
-        lifecycleV2 = LifecycleV2(dataStore: NamedCollectionDataStore(name: name))
+        lifecycleV2 = LifecycleV2(dataStore: NamedCollectionDataStore(name: name), dispatch: MobileCore.dispatch(event:))
```
praveek, line 35: Can we add small comment before `lifecycleState` and `lifecycleV2` declaration on what they handle?

AEPLifecycle/Sources/LifecycleState.swift
```diff
@@ -104,7 +104,7 @@ struct LifecycleState {
 
     /// Pauses the current lifecycle session
     /// - Parameter pauseDate: date at which the pause event occurred
-    mutating func pause(pauseDate: Date) {
+    mutating func pause(pauseDate: Date, dispatchApplicationCloseEvent: ((_ xdm: [String: Any]) -> Void)? = nil ) {
```
praveek, line 107: Can we revert this?

AEPLifecycle/Sources/LifecycleV2/LifecycleV2.swift
```diff
@@ -28,10 +30,12 @@ class LifecycleV2 {
     /// Creates a new `LifecycleV2` with the given `NamedCollectionDataStore`
     ///
     /// - Parameter dataStore: The `NamedCollectionDataStore` used for reading and writing data to persistence
-    init(dataStore: NamedCollectionDataStore) {
+    init(dataStore: NamedCollectionDataStore, metricsBuilder: LifecycleV2MetricsBuilder? = nil, dispatch: @escaping (Event) -> Void) {
```
praveek, line 33: I don't think we need `metricsBuilder` passed into init() function?

AEPLifecycle/Sources/LifecycleV2/LifecycleV2.swift
```diff
@@ -28,10 +30,12 @@ class LifecycleV2 {
     /// Creates a new `LifecycleV2` with the given `NamedCollectionDataStore`
     ///
     /// - Parameter dataStore: The `NamedCollectionDataStore` used for reading and writing data to persistence
```
praveek, line 32: Update function comments to include additional parameters.

AEPLifecycle/Sources/LifecycleV2/LifecycleV2.swift
praveek: Update function comments to include additional parameters.

AEPLifecycle/Sources/LifecycleV2/LifecycleV2.swift
```diff
@@ -53,7 +57,20 @@ class LifecycleV2 {
         stateManager.update(state: .START) { [weak self] (updated: Bool) in
             guard let self = self else { return }
             guard updated else { return }
-            // todo: MOB-14878 create launch application event and if needed the close event
+
+            // detect a possible crash/incorrect start/pause implementation
+            if !isInstall && self.isCloseUnknown(prevAppStart: self.dataStoreCache.getAppStartDate(), prevAppPause: self.dataStoreCache.getAppPauseDate()) {
+                // in case of an unknown close situation, use the last known app close event timestamp
+                guard let xdm = self.xdmMetricsBuilder.buildAppCloseXDMData(launchDate: self.dataStoreCache.getAppStartDate() ?? Date(), closeDate: self.dataStoreCache.getCloseDate() ?? Date(), isCloseUnknown: false) else { return }
```
praveek, line 64: isCloseUnknown: true

praveek: I would prefer changing the logic to 
if let xdm = self.xdmMetricsBuilder.buildAppCloseXDMData(...) { 
 self.dispatchApplicationClose(xdm: xdm)
}
instead of returning early.

emdobrin: +1 and can we use the same for buildAppLaunchXDMData and inside pause

AEPLifecycle/Sources/LifecycleV2/LifecycleV2.swift
```diff
@@ -65,10 +82,52 @@ class LifecycleV2 {
         stateManager.update(state: .PAUSE) { [weak self] (updated: Bool) in
             guard let self = self else { return }
             guard updated else { return }
+            guard let xdm = self.xdmMetricsBuilder.buildAppCloseXDMData(launchDate: self.dataStoreCache.getAppStartDate() ?? Date(), closeDate: pauseDate, isCloseUnknown: false) else { return }
```
emdobrin, line 85: Couple of things that need to be updated here:
- the pauseDate needs to be persisted with dataStoreCache.setAppPauseDate
- buildAppCloseXDMData seems to require optional Date for launchDate and closeDate and handle those appropriately, we should not pass current Date() as app start date in this situation as this impacts the session length computation.

AEPLifecycle/Sources/LifecycleV2/LifecycleV2.swift
emdobrin: Few other updates required here:
- buildAppCloseXDMData should be updated and include the fallbackCloseEventDate (see Android implementation)
- we should not pass current Date() when app start/app pause are nil, we should handle the case accordingly (see Android computeSessionLength where it checks for both start/pause to be valid values).

AEPLifecycle/Sources/LifecycleV2/LifecycleV2.swift
```diff
@@ -65,10 +82,52 @@ class LifecycleV2 {
...
+            // dispatch application close event with xdm data
+            self.dispatchApplicationClose(xdm: xdm)
+        }
+
+    }
+    /// Identifies if the previous session ended due to an incorrect implementation or possible app crash.
+    ///
+    /// - Parameters:
+    ///   - prevAppStart: start timestamp from previous session
+    ///   - prevAppPause: pause timestamp from previous session
```
emdobrin, line 96: nit: timestamp -> date

AEPLifecycle/Sources/LifecycleV2/LifecycleV2.swift
addb: Yes, I can make metricsBuilder public for testing actually.

AEPLifecycle/Sources/LifecycleV2/LifecycleV2.swift
addb: Yes I wanted to discuss this, so we should return nil if nil launch date found?

emdobrin: We can pass nil to buildAppCloseXDMData -> computeAppCloseData. SessionLength should be 0 if it cannot be computed because of a read fail from persistence.

PR #679, opened 2021-08-04
expanding flexibility of in-app message creation

sbenedicadb:
- adding support for message settings
- overhaul fullscreenmessage class to use message settings and build the webview prior to asking the monitor whether the message should show or not (this allows the delegate to save off and use the constructed `WKWebView` if they want to)

### DIFF
--- a/AEPServices/Sources/ui/MessageSettings.swift
+++ b/AEPServices/Sources/ui/MessageSettings.swift
@@ -15,62 +15,62 @@ import UIKit
 
 @objc(AEPMessageSettings)
 public class MessageSettings: NSObject {
-    let parent: Any?
-    private(set) var width: Int?
-    private(set) var height: Int?
-    private(set) var verticalAlign: MessageAlignment = .center
-    private(set) var verticalInset: Int?
-    private(set) var horizontalAlign: MessageAlignment = .center
-    private(set) var horizontalInset: Int?
-    private(set) var uiTakeover: Bool = true
-    private(set) var animate: Bool = true
-    private(set) var gestures: [MessageGesture: URL]?
+    public let parent: Any?
+    public private(set) var width: Int?
+    public private(set) var height: Int?
+    public private(set) var verticalAlign: MessageAlignment = .center
+    public private(set) var verticalInset: Int?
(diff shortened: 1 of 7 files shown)


AEPServices/Sources/ui/fullscreen/FullscreenMessage.swift
```diff
@@ -239,4 +259,121 @@ public class FullscreenMessage: NSObject, FullscreenPresentable {
...
+        }
+
+        return frame
+    }
+
+    // returns the width of the screen, measured in points
+    private var screenWidth: CGFloat {
+        return UIScreen.main.bounds.width
+    }
+
```
PravinPK, line 283: Doc
// returns the height of the screen, measured in points

AEPServices/Sources/ui/fullscreen/FullscreenMessage.swift
```diff
@@ -239,4 +259,121 @@ public class FullscreenMessage: NSObject, FullscreenPresentable {
...
+    // default value is full screen width
+    private var width: CGFloat {
+        if let settingsWidth = settings?.width {
+            return screenWidth * CGFloat(settingsWidth) / 100
+        }
+
+        return screenWidth
+    }
+
+    // height in settings represents a percentage of the screen.
```
PravinPK, line 299: method docs can be formatted with `///` instead of `//`

AEPServices/Sources/ui/MessageSettings.swift
```diff
@@ -0,0 +1,77 @@
...
+ OF ANY KIND, either express or implied. See the License for the specific language
+ governing permissions and limitations under the License.
+ */
+
+import Foundation
+import UIKit
+
+@objc(AEPMessageSettings)
+public class MessageSettings: NSObject {
+    public let parent: Any?
```
praveek, line 18: I am not familiar with the use case here but does `parent` change based on when the message is displayed?

sbenedicadb: no, the parent is the class that owns the Message that's being created - it's just a dependency injection to allow more control for the caller

PR #686, opened 2021-08-10
Limit number of times a dispatch consequence event can be processed

kevinlind:
## Description
Limit the number of times a dispatch consequence may be processed by the rules engine. The issue is a dispatch consequence rule may be the tigger for an another dispatch consequence rule, causing an infinite loop of dispatched events. This PR adds a map to track the ID of dispatched events from the rule consequence and then aborts the processing of a dispatch consequence if the triggering event appears in the map with too high of a count.

Internal ticket: MOB-14753

## Types of changes

## Checklist:


AEPCore/Sources/rules/LaunchRulesEngine.swift
```diff
@@ -104,6 +107,7 @@ public class LaunchRulesEngine {
     }
 
     private func evaluateRules(for event: Event) -> Event {
+        let dispatchChainCount = dispatchChainedEventsCount.removeValue(forKey: event.id)
```
emdobrin, line 110: The id should not be removed until it is evaluated and the chain needs to be updated.

AEPCore/Sources/rules/LaunchRulesEngine.swift
```diff
@@ -128,12 +132,19 @@ public class LaunchRulesEngine {
                     eventData = modifiedEventData
 
                 case LaunchRulesEngine.CONSEQUENCE_TYPE_DISPATCH:
+                    if let unwrappedDispatchCount = dispatchChainCount, unwrappedDispatchCount > LaunchRulesEngine.MAX_CHAINED_CONSEQUENCE_COUNT {
```
emdobrin, line 135: Is the intention to let this trigger the dispatch consequence twice or should the condition be unwrappedDispatchCount >= LaunchRulesEngine.MAX_CHAINED_CONSEQUENCE_COUNT?

AEPCore/Sources/rules/LaunchRulesEngine.swift
kevinlind: I remove the id here so the list gets cleaned up in cases where the event does not trigger the dispatch consequence event. In cases where no rules matched the event or the event triggers an add or modify data consequence, the id should get removed as the event was evaluated.  It is my understanding that events are evaluated for rules once, so once the `evaluateRules` function completes, the event id no longer needs to be tracked.

AEPCore/Sources/rules/LaunchRulesEngine.swift
kevinlind: We can do it either way. What is more readable to you? 
The max count is really arbitrary, so if you prefer to use `>=` and set the max count to 2, that's fine.
Also, is 2 iterations a good max count or is it too conservative?  There could be valid use cases to tigger a rule off of a dispatched event. If we are using this check as a failsafe to prevent an infinite loop, then we should set the count higher. If we want to prevent any chaining of events dispatched from rules, then we can set the count to 1 or 2.

emdobrin: I find it more intuitive with `>= MAX_COUNT`. I think we can start with 1 and relax this later if needed.

AEPCore/Sources/rules/LaunchRulesEngine.swift
emdobrin: This seems fair except for the case where an event is created and then dispatched multiple times, the Event UUID does not change in this case.

kevinlind: In what cases would an event get dispatched multiple times? It's possible for anyone to dispatch a received event, but I'm not familiar with where in the SDK that occurs.

emdobrin: 💭  This scenario may cause some other interesting behavior in the EventHub, if multiple events are queued consecutively the last one will override the event number of the previous one before.

emdobrin: Looks like this is not prevented today, so you could have something like:
```
/// some pre-computed event
let event: Event = Event(name: "test", type: "test.type", source: "test.source", data: ["some":"data"])
...
func click() {
    MobileCore.dispatch(event: event)
}
```

kevinlind: Remember this list only contains the UUID of events dispatched from this Rules Engine, so we'd need someplace else to be re-dispatching these events to the Event Hub for them to be evaluated multiple times. In your scenario, events being evaluated will not override the event UUIDs in the list, but it will cause the count set to new events created from the dispatch consequence to be set to 0.

kevinlind: A compromise would be to keep the triggering event UUID in the list if the event matches a dispatch consequence. If the the event does not match a dispatch consequence, it is removed from the list. This will mean events which match a dispatch consequence rule will always stay in the list. As this is an unexpected case, it is assumed it will be a small number of events.
One question here is if the event is evaluated a second time, should the dispatch consequence get processed again? On the one hand, the consequence was already processed when the event as first received. On the other hand, we'd probably want the SDK to behave the same if the same event is dispatched multiple times, which would include processing the same rules the event matches.

AEPCore/Sources/rules/LaunchRulesEngine.swift
```diff
@@ -132,9 +132,17 @@ public class LaunchRulesEngine {
...
                 case LaunchRulesEngine.CONSEQUENCE_TYPE_DISPATCH:
-                    if let unwrappedDispatchCount = dispatchChainCount, unwrappedDispatchCount >= LaunchRulesEngine.MAX_CHAINED_CONSEQUENCE_COUNT {
-                        Log.trace(label: LOG_TAG, "(\(self.name)) : Unable to process a DispatchConsequence Event, max dispatch consequences met for this event, \(event)")
-                        continue
+
+                    if let unwrappedDispatchCount = dispatchChainCount {
+                        // Keep track of triggering event in edge case of event being dispatched multiple times
+                        // This only applies to events created from this dispatch consequence, hence the check
+                        // for a null 'dispatchChainCount' prior to adding back to dictionary
+                        dispatchChainedEventsCount[event.id] = LaunchRulesEngine.MAX_CHAINED_CONSEQUENCE_COUNT
```
praveek, line 140: Do we need this logic? This means irrespective of the `LaunchRulesEngine.MAX_CHAINED_CONSEQUENCE_COUNT` the chain length can not be more than 2.

AEPCore/Sources/rules/LaunchRulesEngine.swift
praveek: I feel calling `MobileCore.dispatch(event: event)` should behave the same way and these checks were to limit the events automatically dispatched by rules. We should create a separate task and look into handling dispatching same event multiple times.

emdobrin: @praveek added https://github.com/adobe/aepsdk-core-ios/issues/688 for this.

AEPCore/Sources/rules/LaunchRulesEngine.swift
kevinlind: This means that a "link" in the chain will not get processed more than once. The chain length will still be MAX_CHAINED_CONSEQUENCE_COUNT. This is catching the case where an event dispatched by this consequence will only trigger a dispatch consequence once. If the same event is reprocessed, then it will not dispatch more events. As I'm checking if `dispatchChainCount` is not null, only events which were originally created by the dispatch consequence are handled, but not events which originate else where.

AEPCore/Sources/rules/LaunchRulesEngine.swift
```diff
@@ -128,12 +132,27 @@ public class LaunchRulesEngine {
...
                 case LaunchRulesEngine.CONSEQUENCE_TYPE_DISPATCH:
+
+                    if let unwrappedDispatchCount = dispatchChainCount {
+                        // Keep track of triggering event in edge case of event being dispatched multiple times
+                        // This only applies to events created from this dispatch consequence, hence the check
+                        // for a null 'dispatchChainCount' prior to adding back to dictionary
+                        dispatchChainedEventsCount[event.id] = LaunchRulesEngine.MAX_CHAINED_CONSEQUENCE_COUNT
+
+                        if unwrappedDispatchCount >= LaunchRulesEngine.MAX_CHAINED_CONSEQUENCE_COUNT {
+                            Log.trace(label: LOG_TAG, "(\(self.name)) : Unable to process a DispatchConsequence Event, max dispatch consequences met for this event, \(event)")
```
emdobrin, line 143: Rewording suggestion: `Unable to process dispatch consequence, max chained dispatch consequences limit of \(MAX_CHAINED_CONSEQUENCE_COUNT) met for this event uuid \(event.id)`

AEPCore/Tests/FunctionalTests/RulesEngineFunctionalTests.swift
```diff
@@ -871,6 +872,367 @@ class RulesEngineFunctionalTests: XCTestCase {
...
+        //              "values": [
+        //                "yes"
+        //              ]
+        //            }
+        //          }
+        //        ]
+        //    ---------- dispatch event rule 2 consequence ----------
+        //        "detail": {
+        //           "type" : "com.adobe.eventType.edge",
+        //           "source" : "com.adobe.eventSource.requestContent",
```
emdobrin, line 1203: ❓ do we want to keep these Edge/Lifecycle test cases in AEPCore RulesEngineFunctionalTests or should we make these tests more generic instead and cover these as integration tests

AEPCore/Sources/rules/LaunchRulesEngine.swift
emdobrin: This becomes little more complicated than the use-case we started this thread of changes with. Can we instead keep the count for the original chain link for this triggering event? 
```
e.g. MAX_CHAINED_CONSEQUENCE_COUNT = 2
Let's assume E1 triggered two dispatch consequences E2 and E3, E2 triggered E4 
dispatchChainedEventsCount should look like:
uuid E2: 1
uuid E3: 1
uuid E4: 2
E4 cannot trigger new dispatch consequences as its chain maxed out, 
but a replay of E2 and E3 is allowed and they will generate at most one additional event each.
```

AEPCore/Tests/FunctionalTests/RulesEngineFunctionalTests.swift
kevinlind: I like having these functional tests as they run fast and it's easier to run negative tests than with integration tests. I've just added a few integration tests as a sanity check to verify all the end-to-end plumbing is hooked up.

AEPCore/Sources/rules/LaunchRulesEngine.swift
kevinlind: I feel allowing a tree structure like that is more confusing. It's not clear why E2 can be replayed while E4 cannot. 

For example, if MAX_CHAINED_CONSEQUENCE_COUNT = 5, and there are rules which chain dispatch consequences:
E0 -> E1 (1) -> E2 (2) -> E3 (3) -> E4 (4) -> E5 (5)
A replay of E1 produces:
E1 (2) -> E2a (3) -> E3a (4) -> E4a (5)
Another replay of E1 produces:
E1 (3) -> E2b (4) -> E3b (5)
Etc.

This behavior is not intuitive as users have little in site into the internal count we assign to each event. What I'm advocating for is preventing the replay of events created from the dispatch consequence.

praveek: I feel we should use this count to only limits events automatically generated by rules in response to a single `MobileCore.dispatch` API call. Let us treat the replayed events as new events to make the behaviour predictable. 
We should take a second pass when we look into #688

AEPIntegrationTests/resources/rules_rules.json
```diff
@@ -0,0 +1,212 @@
+{
```
praveek, line 1: Can we rename this file to something better?

AEPCore/Sources/rules/LaunchRulesEngine.swift
```diff
@@ -128,12 +132,20 @@ public class LaunchRulesEngine {
...
+                        Log.trace(label: LOG_TAG, "(\(self.name)) : Unable to process dispatch consequence, max chained dispatch consequences limit of \(LaunchRulesEngine.MAX_CHAINED_CONSEQUENCE_COUNT) met for this event uuid \(event.id)")
+                        continue
+                    }
                     guard let dispatchEvent = processDispatchConsequence(consequence: consequenceWithConcreteValue, eventData: eventData)  else {
                         continue
                     }
                     Log.trace(label: LOG_TAG, "(\(self.name)) : Generating new dispatch consequence result event \(dispatchEvent)")
                     extensionRuntime.dispatch(event: dispatchEvent)
 
+                    // Keep track of dispatch consequence events to prevent trigging of infinite dispatch consequences
```
emdobrin, line 146: typo: triggering

emdobrin (approved):

AEPCore/Sources/rules/LaunchRulesEngine.swift
emdobrin: @praveek agreed, will simplify this logic in Android too.

praveek (approved):

Issue #668, opened 2021-07-22
Test app for Core extensions


emdobrin:
Create a test app that includes all Core extensions that can be used for manual testing during development. This test app can include detailed or edge case test scenarios.

PR #699, opened 2021-08-19
[MOB-14763] Test app for Core (Lifecycle / Assurance support for now)

cdhoffmann:
A simple test app with support for Assurance use and Lifecycle Edge in order to perform manual testing. Skeleton code has been setup for multiple views support for Core extensions as well as per [this issue](https://github.com/adobe/aepsdk-core-ios/issues/668)


emdobrin:
@cdhoffmann can you please include some screenshots in the PR description?

TestApps/TestApp_Swift/CoreView.swift
```diff
@@ -0,0 +1,21 @@
+//
+//  CoreView.swift
+//  TestApp_Swift
+//
+//  Created by Christopher Hoffman on 8/19/21.
+//  Copyright © 2021 Adobe. All rights reserved.
+//
```
emdobrin, line 7: requires ©️ update - same for the other files

TestApps/TestApp_Swift/MenuView.swift
```diff
@@ -0,0 +1,43 @@
...
+    var body: some View {
+        NavigationView {
+            Form {
+                Section(header: Text("Core Extensions")) {
+                    NavigationLink(destination: CoreView().navigationBarTitle("Core")) {
+                        Text("Core")
+                    }
+
+                    NavigationLink(destination: IdentityView().navigationBarTitle("Identity")) {
+                        Text("Identity")
```
emdobrin, line 24: I am thinking we should also add a standalone view for Lifecycle for various tests - can we add a Start and Pause buttons and allow to include additional data on the pause API call?

emdobrin (commented):
It would be good to also include this app as target in the AEPCore.xcworkspace so you write code & test at the click of a button.

TestApps/TestApp_Swift/MenuView.swift
cdhoffmann: Good idea. Should i remove the start / pause calls from the sceneDelegate and appDelegate to have complete control over when the APIs are called? Or leave them in there to follow the standard, and just add the buttons in addition?

TestApps/TestApp_Swift/AppDelegate.swift
```diff
@@ -0,0 +1,58 @@
...
+
+@main
+class AppDelegate: UIResponder, UIApplicationDelegate {
+
+    private let LAUNCH_ENVIRONMENT_FILE_ID = ""
+
+    func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
+        // Override point for customization after application launch.
+        let appState = application.applicationState
+        let extensions = [AEPIdentity.Identity.self,
```
praveek, line 28: nit: Any reason it is not just Identity?

TestApps/TestApp_Swift/AppDelegate.swift
```diff
@@ -0,0 +1,58 @@
...
+                          Lifecycle.self,
+                          Signal.self,
+                          Assurance.self]
+
+        MobileCore.registerExtensions(extensions, {
+            MobileCore.configureWith(appId: self.LAUNCH_ENVIRONMENT_FILE_ID)
+            if appState != .background {
```
praveek, line 35: We can remove this logic and allow full control of start and pause APIs in the Lifecycle view.

PR #715, opened 2021-09-17
starting work on event history

sbenedicadb:


AEPCore/Sources/eventhub/Event.swift
```diff
@@ -39,28 +39,38 @@ public class Event: NSObject, Codable {
...
     /// - Parameters:
     ///   - name: Name for the `Event`
     ///   - type: `EventType` for the `Event`
     ///   - source: `EventSource` for the `Event`
     ///   - data: Any associated data with this `Event`
+    ///   - mask: Defines which properties should be used in creation of the Event's hash
     @objc
-    public convenience init(name: String, type: String, source: String, data: [String: Any]?) {
-        self.init(name: name, type: type, source: source, data: data, requestEvent: nil)
+    public convenience init(name: String, type: String, source: String, data: [String: Any]?, mask: [String]? = nil) {
```
yangyansong-adbe, line 72: I think it might be a breaking change for Core ABIs. 
Should we add ad new convenience initializer instead of modifying this one?

AEPCore/Sources/eventhub/EventHub.swift
```diff
@@ -30,6 +30,7 @@ final class EventHub {
     private let eventQueue = OperationOrderer<Event>("EventHub")
     private var preprocessors = ThreadSafeArray<EventPreprocessor>(identifier: "com.adobe.eventHub.preprocessors.queue")
     private var started = false // true if the `EventHub` is started, false otherwise. Should only be accessed from within the `eventHubQueue`
+    private var eventHistory = EventHistory()
```
yangyansong-adbe, line 33: How are we going to pass the `EventHistory` instance to `RulesEngine`s? I guess we need to add a public getter in `ExtensionContainer`/`ExtensionRuntime`, which will be passed to the RulesEngine instance here (https://github.com/adobe/aepsdk-core-ios/blob/14dbeef70cb1375ea86667db5571e74522ccc23f/AEPCore/Sources/rules/LaunchRulesEngine.swift#L51), right?
If so, can we also make those changes in this PR?

AEPCore/Sources/eventhub/history/EventData+FNV1A32.swift
```diff
@@ -0,0 +1,40 @@
...
+ Unless required by applicable law or agreed to in writing, software distributed under
+ the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ OF ANY KIND, either express or implied. See the License for the specific language
+ governing permissions and limitations under the License.
+ */
+
+import Foundation
+
+extension Dictionary where Key == String, Value == Any {
+    /// flattens the dictionary then calls inner
```
yangyansong-adbe, line 16: Nit: This comment is not clear enough for its callers.  Can we give more detailed information here?

AEPCore/Sources/eventhub/history/EventData+FNV1A32.swift
```diff
@@ -0,0 +1,40 @@
...
+        if let mask = mask {
+            alphabeticalKeys = self.keys.filter({ mask.contains($0) }).sorted()
+        } else {
+            alphabeticalKeys = self.keys.sorted()
+        }
+
+        var hash: UInt32 = 0
+        for i in 0..<alphabeticalKeys.count {
+            let key = alphabeticalKeys[i]
+            let kvpString = "\(key):\(self[key] ?? "")"
```
yangyansong-adbe, line 34: Just a reminder: in Swift, the default `DefaultStringInterpolation
` (https://developer.apple.com/documentation/swift/defaultstringinterpolation) could be overridden.

AEPCore/Sources/eventhub/history/EventHistory.swift
```diff
@@ -0,0 +1,81 @@
...
+
+ Unless required by applicable law or agreed to in writing, software distributed under
+ the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ OF ANY KIND, either express or implied. See the License for the specific language
+ governing permissions and limitations under the License.
+ */
+
+import Foundation
+
+public class EventHistory {
```
yangyansong-adbe, line 15: I think we don't need to make it a public class as we only need to use it in LaunchRulesEngine class.

AEPCore/Sources/eventhub/history/EventHistory.swift
```diff
@@ -0,0 +1,81 @@
...
+
+    func recordEvent(_ event: Event, handler: ((Bool) -> Void)? = nil) {
+        guard let hash = event.data?.fnv1a32(mask: event.mask) else {
+            return
+        }
+
+        db.insert(hash: hash, handler: handler)
+    }
+
+    func getEvents(_ events: [EventHistoryRequest], enforceOrder: Bool, handler: @escaping ([EventHistoryResult]) -> Void) {
```
yangyansong-adbe, line 34: events: [EventHistoryRequest]   -> requests: [EventHistoryRequest]

AEPCore/Sources/eventhub/EventHub.swift
```diff
@@ -86,6 +87,11 @@ final class EventHub {
             self?.eventQueue.add(event)
             Log.trace(label: self?.LOG_TAG ?? "EventHub",
                       "Dispatching Event #\(String(describing: self?.eventNumberMap[event.id] ?? 0)) - \(event)")
+
+            // record the event in history if it has a mask
+            if event.mask != nil {
+                self?.eventHistory?.recordEvent(event)
```
yangyansong-adbe, line 93: We'd better add a debug log here if the `EventHistory` is not initialized correctly.

AEPCore/Sources/eventhub/history/EventHistory.swift
```diff
@@ -0,0 +1,81 @@
...
+        var results: [EventHistoryResult] = []
+
+        if enforceOrder {
+            var previousEventOldestOccurrence: Date?
+            for event in events {
+                let eventHash = event.mask.fnv1a32()
+                let from = previousEventOldestOccurrence ?? event.fromDate
+                db.select(hash: eventHash, from: from, to: event.toDate) { result in
```
yangyansong-adbe, line 42: We'd better optimize the database operations here not to open/close the database per `request`.  (to open the database connection is a time-consuming operation which may re-read the database structure)

AEPCore/Sources/eventhub/history/EventData+FNV1A32.swift
sbenedicadb: it's a good point @yangyansong-adbe.  we can re-visit to ensure this is consistent.

PR #726, opened 2021-11-02
-update to use new function operand from rules engine

sbenedicadb:


AEPCore/Sources/rules/JSONRulesParser.swift
```diff
@@ -163,36 +163,46 @@ class JSONCondition: Codable {
             EventHistoryRequest(mask: $0, from: fromDate, to: toDate)
         })
 
-        let historyOperand = Operand<Int>(function: {
-            var returnValue: Int = 0
-            let semaphore = DispatchSemaphore(value: 0)
+        let params: [Any] = [runtime, requestEvents, searchType!]
```
praveek, line 166: Do we need to pass `searchType!` instead of `searchType`?

praveek (approved):

AEPCore/Sources/rules/JSONRulesParser.swift
```diff
@@ -163,36 +163,46 @@ class JSONCondition: Codable {
...
+        let historyOperand = Operand<Int>(function: getHistoricalEventCount, parameters: params)
```
praveek, line 167: This capture a reference to `self`. I don't see any issue with it but given these classes are created by JSONDecoder do you think we should move this to be a top level / static function?

PR #730, opened 2021-11-05
Fix a bug in persisting Date object

praveek:
## Description

JSON Encoder shipped as part of Swift standard library in iOS, tvOS versions < 13 fails to encode top level fragments. Persist date as double in iOS, tvOS versions < 13. 

## Types of changes

## Checklist:

### DIFF
--- a/AEPServices/Sources/storage/NamedCollectionDataStore.swift
+++ b/AEPServices/Sources/storage/NamedCollectionDataStore.swift
@@ -168,12 +168,33 @@ public class NamedCollectionDataStore {
     }
 
     public func setObject<T: Codable>(key: String, value: T) {
+        // https://bugs.swift.org/browse/SR-6163
+        // JSON Encoder shipped as part of Swift standard library in iOS versions < 13 fails to encode top level fragments. Persist date as double in iOS versions < 13
+        if T.self == Date.self, let value = value as? Date {
+            guard #available(iOS 13.0, tvOS 13.0, *) else {
+                set(key: key, value: value.timeIntervalSince1970)
+                return
+            }
+        }
+
         let encoder = JSONEncoder()
         let encodedValue = try? encoder.encode(value)
         set(key: key, value: encodedValue)
     }
 
     public func getObject<T: Codable>(key: String, fallback: T? = nil) -> T? {
+        // https://bugs.swift.org/browse/SR-6163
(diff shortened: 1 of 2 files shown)


emdobrin (commented):
This should also cover the use-case encode as double < iOS 13, decode as Date > iOS 13 for OS upgrade scenario, let's check on that too

cdhoffmann:
Can we add a test to cover this as well?

cdhoffmann (approved):

AEPServices/Tests/services/NamedCollectionDataStoreTest.swift
```diff
@@ -385,6 +385,34 @@ class NamedCollectionDataStoreTest: XCTestCase {
...
+
+        XCTAssertTrue(mockKeyValueService.setCalled)
+        if #available(iOS 13, tvOS 13, *) {
+            let encodedDate = try? JSONEncoder().encode(date)
+            XCTAssertEqual(mockKeyValueService.setValue as? Data, encodedDate)
+        } else {
+            XCTAssertEqual(mockKeyValueService.setValue as? Double, date.timeIntervalSince1970)
+        }
+    }
+    func testGetDateCodable() {
```
emdobrin, line 400: can we also add a test where we mock the store as double and read as date?

praveek: The next test case does it.

praveek:
@emdobrin Good catch with upgrade scenario. I have fixed the issue and updated tests.

AEPServices/Sources/storage/NamedCollectionDataStore.swift
```diff
@@ -168,12 +168,27 @@ public class NamedCollectionDataStore {
...
+            }
+        }
+
         let encoder = JSONEncoder()
         let encodedValue = try? encoder.encode(value)
         set(key: key, value: encodedValue)
     }
 
     public func getObject<T: Codable>(key: String, fallback: T? = nil) -> T? {
+        // We persist date as double in iOS versions < 13.
```
emdobrin, line 186: rewording suggestion instead of we persist: `setObject persists Date values as double in iOS version < 13`
I think we can also add a `@see setObject`  reference in the API docs

emdobrin (approved):

PR #738, opened 2021-11-10
-add messaging event type and sources

sbenedicadb:


AEPCore/Sources/eventhub/EventSource.swift
```diff
@@ -34,4 +34,5 @@ public class EventSource: NSObject {
     public static let resetComplete = "com.adobe.eventSource.resetComplete"
     public static let applicationLaunch = "com.adobe.eventSource.applicationLaunch"
     public static let applicationClose = "com.adobe.eventSource.applicationClose"
+    public static let personalizationDecisions = "personalization:decisions"
```
praveek, line 37: Any reason we are moving away from `com.adobe.eventSource.` prefix? It will be consistent if we name it "com.adobe.eventSource.personalizationDecisions".

I also feel we should stop adding eventSource/eventType constants in Core unless they are very standard and could be reused by multiple extensions. Thoughts?

PR #741, opened 2021-11-13
[MOB-14668] Wait for Analytics shared state when registered for GetUrlVariables and AppendToUrl APIs

addb:
## Types of changes

## Checklist:

### DIFF
--- a/AEPIdentity/Sources/Identity.swift
+++ b/AEPIdentity/Sources/Identity.swift
@@ -62,6 +62,10 @@ import Foundation
             return state?.readyForSyncIdentifiers(event: event, configurationSharedState: configSharedState) ?? false
         } else if event.type == EventType.configuration, event.source == EventSource.requestIdentity {
             return MobileIdentities().areSharedStatesReady(event: event, sharedStateProvider: getSharedState(extensionName:event:))
+        } else if event.type == EventType.identity, event.source == EventSource.requestIdentity {
+            guard getSharedState(extensionName: IdentityConstants.Hub.SHARED_OWNER_NAME, event: event)?.status ?? .none != .pending else {
+                return false
+            }
         }
 
         return getSharedState(extensionName: IdentityConstants.SharedStateKeys.CONFIGURATION, event: event)?.status == .set
@@ -180,7 +184,14 @@ import Foundation
     private func processAppendToUrl(baseUrl: String, event: Event) {
         guard let properties = state?.identityProperties else { return }
         guard let configurationSharedState = getSharedState(extensionName: IdentityConstants.SharedStateKeys.CONFIGURATION, event: event)?.value else { return }
-        let analyticsSharedState = getSharedState(extensionName: "com.adobe.module.analytics", event: event)?.value ?? [:]
+
+        let analyticsRegistered = isExtensionRegistered(name: IdentityConstants.SharedStateKeys.ANALYTICS, event: event)
+
+        if analyticsRegistered && getSharedState(extensionName: IdentityConstants.SharedStateKeys.ANALYTICS, event: event)?.status == .pending {
+            return
+        }
+
+        let analyticsSharedState = getSharedState(extensionName: IdentityConstants.SharedStateKeys.ANALYTICS, event: event)?.value ?? [:]
         let updatedUrl = URLAppender.appendVisitorInfo(baseUrl: baseUrl, configSharedState: configurationSharedState, analyticsSharedState: analyticsSharedState, identityProperties: properties)
 
         // dispatch identity response event with updated url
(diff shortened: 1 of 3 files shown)


AEPIdentity/Sources/Identity.swift
```diff
@@ -180,7 +184,14 @@ import Foundation
     private func processAppendToUrl(baseUrl: String, event: Event) {
         guard let properties = state?.identityProperties else { return }
         guard let configurationSharedState = getSharedState(extensionName: IdentityConstants.SharedStateKeys.CONFIGURATION, event: event)?.value else { return }
-        let analyticsSharedState = getSharedState(extensionName: "com.adobe.module.analytics", event: event)?.value ?? [:]
+
+        let analyticsRegistered = isExtensionRegistered(name: IdentityConstants.SharedStateKeys.ANALYTICS, event: event)
+
+        if analyticsRegistered && getSharedState(extensionName: IdentityConstants.SharedStateKeys.ANALYTICS, event: event)?.status == .pending {
+            return
```
kevinlind, line 191: I think all this work needs to be done in `readyForEvent`. Returning from the function simply stops processing the event, but it won't keep it in the queue. Additionally, without a dispatched response, the public API will simply time out.

AEPIdentity/Sources/Identity.swift
```diff
@@ -62,6 +62,10 @@ import Foundation
             return state?.readyForSyncIdentifiers(event: event, configurationSharedState: configSharedState) ?? false
         } else if event.type == EventType.configuration, event.source == EventSource.requestIdentity {
             return MobileIdentities().areSharedStatesReady(event: event, sharedStateProvider: getSharedState(extensionName:event:))
+        } else if event.type == EventType.identity, event.source == EventSource.requestIdentity {
+            guard getSharedState(extensionName: IdentityConstants.Hub.SHARED_OWNER_NAME, event: event)?.status ?? .none != .pending else {
```
kevinlind, line 66: I believe all the dependency checking needs to happen here as returning `false` will let the Event Hub know the extension isn't able to process the event. So you'll need to check if Analytics is registered and check the if the Analytics shared state is pending here.
Also check the event for either `baseUrl` or `urlVariables` so the check only applies to `AppendToUrl` and `GetUrlVariables` requests.

kevinlind (changes requested):

AEPIdentity/Sources/Identity.swift
```diff
@@ -62,6 +62,19 @@ import Foundation
             return state?.readyForSyncIdentifiers(event: event, configurationSharedState: configSharedState) ?? false
         } else if event.type == EventType.configuration, event.source == EventSource.requestIdentity {
             return MobileIdentities().areSharedStatesReady(event: event, sharedStateProvider: getSharedState(extensionName:event:))
+        } else if event.type == EventType.identity, event.source == EventSource.requestIdentity, ( event.baseUrl != nil ||  event.urlVariables ) {
+            // check for hub sharedstate info and analytics registration info only if the Analytics shared state is not already set
+            if getSharedState(extensionName: IdentityConstants.SharedStateKeys.ANALYTICS, event: event)?.status != .set {
```
kevinlind, line 67: From the code for `getSharedState`, it appears if the extension is not registered it will return nil. Can you check with Core team to verify if we can rely on that to determine if an extension is registered or not?
https://github.com/adobe/aepsdk-core-ios/blob/main/AEPCore/Sources/eventhub/EventHub.swift#L230
If that is the case then I think you could change this to something like 
```
if let analyticsSharedState = getSharedState(extensionName: IdentityConstants.SharedStateKeys.Analytics, event: event), analyticsSharedState.status != .set { return false }
```

AEPIdentity/Tests/IdentityTests.swift
```diff
@@ -486,4 +486,49 @@ class IdentityTests: XCTestCase {
         // expect true since identity has booted
         XCTAssertTrue(identity.readyForEvent(firstEvent))
     }
+
+    // analytics shared state not available and hub shared state pending
```
kevinlind, line 490: Can you add tests for getUrlVariables events.
Also, add tests for AppendToUrl and getUrlVariables where Analytics is not registered, there is no shared state and `readyForEvent` returns true.

AEPIdentity/Sources/Identity.swift
kevinlind: If you cannot simplify this check with the above, can you reduce the number of calls to `getSharedState` to just two? Once for Analytics and once for Event Hub?  Looks like currently there are two calls for Analytics and two for Event Hub.

kevinlind (approved):

Issue #714, opened 2021-09-17
Deprecate getApplicationVersion function


praveek:
SystemInfoService exposes following functions 
```
// Returns CFBundleVersion string (Build number in Xcode)
func getApplicationBuildNumber() -> String?

// Returns CFBundleVersion string (Build number in Xcode)
func getApplicationVersion() -> String?

// Returns CFBundleShortVersionString string (Version number in Xcode)
func getApplicationVersionNumber() -> String?
```
We should go ahead and deprecate `getApplicationVersion()` as it is duplicate. It also confusing to have two functions with similar names (`getApplicationVersion` and `getApplicationVersionNumber`)

PR #742, opened 2021-11-22
Deprecate get application version

cdhoffmann:
Marks getApplicationVersion as deprecated. We still need to come up with proper messaging for when it will be removed. #714


AEPServices/Sources/SystemInfoService.swift
```diff
@@ -89,6 +89,8 @@ public protocol SystemInfoService {
 
     /// Returns the application version.
     /// - Return: `String` Application version
+    // TODO: - Include planned deprecation version in message
+    @available(*, deprecated, message: "Please use getApplicationVersionNumber instead")
```
praveek, line 93: Can you also include rename: "getApplicationVersionNumber"?  XCode gives a hint to replace it with new function

praveek (approved):

Issue #508, opened 2021-01-15
random failure of the test testUnRegisterExtensionsSimple 


shalehaha:

shalehaha:
testRegisterEventListenerMultipleLisenersForSameEvents

shalehaha:
testRegisterEventListener

PR #743, opened 2021-11-22
fix unregisterExtensionSimple

cdhoffmann:
Fixes an intermittent failure due to a race condition. #508


AEPCore/Tests/MobileCoreTests.swift
```diff
@@ -107,14 +107,13 @@ class MobileCoreTests: XCTestCase {
         let expectation = XCTestExpectation(description: "unregistration completed in timely fashion")
         expectation.assertForOverFulfill = true
         MockExtension.unregistrationClosure = { expectation.fulfill() }
-        MobileCore.registerExtensions([MockExtension.self])
-
-        // test
-        MobileCore.unregisterExtension(MockExtension.self)
+        MobileCore.registerExtensions([MockExtension.self]) {
```
praveek, line 110: This could again cause a failure if the test completes before registerExtensions finishes execution.
Can you modify as below to avoid race condition
```
let semaphore = DispatchSemaphore(value: 0)
MobileCore.registerExtensions([MockExtension.self]) {
    semaphore.signal()
}
semaphore.wait()

// test
MobileCore.unregisterExtension(MockExtension.self)
```

praveek (approved):

PR #745, opened 2021-11-30
Add pod install to archive command

cdhoffmann:
Add pod install to archive command and remove the readme instructions to run pod install before make archive.


Makefile
```diff
@@ -60,6 +60,7 @@ integration-test:
 
 
 archive:
+	pod install
```
praveek, line 63: Add a separate make target `pod-install` and let archive depend on it. 
```
pod-install: 
      pod install

archive: pod-install
      ...
```

cdhoffmann: Got it. Is a --repo-update required?

praveek: Yes. We can add it so that pod install does not fail if the spec repo has not been updated.

praveek (approved):

PR #750, opened 2021-12-14
-fixing a couple of bugs in fullscreen message logic

sbenedicadb:
- wrong logic when deciding which path to take depending on whether local image is used or not. 
- prevent lossy conversions when writing temp html file


cdhoffmann (approved):
Looks good. Thanks for the fix.

Issue #749, opened 2021-12-03
Fix issues with configureWithAppId


cdhoffmann:
DoD : 

Remove calling configureWithAppId in response to Lifecycle::ResponseContent 

configureWithAppId should clear the persisted appId when passing empty or null string.

PR #751, opened 2021-12-14
ConfigureWithAppId Changes

cdhoffmann:
There are two changes being introduced as per #749.
1. We are no longer calling configureWithAppId in response to Lifecycle::ResponseContent. The reasoning for this change can be found here:[Configuration Network Findings](https://wiki.corp.adobe.com/pages/viewpage.action?spaceKey=~spujari&title=Configuration+Android+extension+testing%3A+Behavior+in+offline+scenarios). Essentially, this was overriding the set local file config unexpectedly.
2. We are now removing app id from persistence when an empty app id is passed into the API as requested by customers.

### DIFF
--- a/AEPCore/Sources/configuration/Configuration.swift
+++ b/AEPCore/Sources/configuration/Configuration.swift
@@ -44,7 +44,6 @@ class Configuration: NSObject, Extension {
         registerPreprocessor(rulesEngine.process(event:))
 
         registerListener(type: EventType.configuration, source: EventSource.requestContent, listener: receiveConfigurationRequest(event:))
-        registerListener(type: EventType.lifecycle, source: EventSource.responseContent, listener: receiveLifecycleResponse(event:))
 
         // If we have an appId stored in persistence, kick off the configureWithAppId event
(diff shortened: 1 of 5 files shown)


AEPCore/Sources/configuration/Configuration.swift
```diff
@@ -44,7 +44,6 @@ class Configuration: NSObject, Extension {
         registerPreprocessor(rulesEngine.process(event:))
 
         registerListener(type: EventType.configuration, source: EventSource.requestContent, listener: receiveConfigurationRequest(event:))
-        registerListener(type: EventType.lifecycle, source: EventSource.responseContent, listener: receiveLifecycleResponse(event:))
```
praveek, line 47: Can you also remove receiveLifecycleResponse function and related code?

praveek (changes requested):

AEPCore/Sources/configuration/Configuration.swift
cdhoffmann: Oops, thought i did. Thanks, done.

praveek (approved):

PR #753, opened 2021-12-20
Fix Identity.getIdentifiers API to return Identifiers list instead of AEPError

kevinlind:
## Description
Identity.getIdentifiers was attempting to cast the list of identifiers from the response event to an Identifiers list, which failed as the response event is encoded as [String: Any]. Therefore getIdentifiers always returned an AEPError. This fix gets the list of identifiers as [String: Any] then builds a list of Identifiers to return to the callback function. Also, if the response event does not contain a visitor identity entry, the API will return an empty list instead of an AEPError.

Internal ticket: MOB-15688

## Types of changes

## Checklist:

### DIFF
--- a/AEPIdentity/Sources/Identity+PublicAPI.swift
+++ b/AEPIdentity/Sources/Identity+PublicAPI.swift
@@ -54,7 +54,7 @@ import Foundation
                 return
             }
 
-            guard var eventData: [String: Any] = responseEvent.data else {
+            guard let eventData: [String: Any] = responseEvent.data else {
                 completion(nil, AEPError.unexpected)
                 return
             }
--- a/AEPIdentity/Sources/Identity+PublicAPI.swift
+++ b/AEPIdentity/Sources/Identity+PublicAPI.swift
@@ -53,23 +53,23 @@ import Foundation
                 completion(nil, AEPError.callbackTimeout)
                 return
             }
-            
+
             guard var eventData: [String: Any] = responseEvent.data else {
                 completion(nil, AEPError.unexpected)
                 return
             }
-            
+
             // return empty list if no Customer Identifiers in response data
             if eventData[IdentityConstants.EventDataKeys.VISITOR_IDS_LIST] == nil {
                 completion([], nil)
                 return
             }
-            
+
             guard let identifiersDict = eventData[IdentityConstants.EventDataKeys.VISITOR_IDS_LIST] as? [[String: Any]] else {
                 completion(nil, AEPError.unexpected)
                 return
             }
-            
+
             let identifiers = identifiersDict.map { CustomIdentity.from(dict: $0) }.compactMap { $0 }
 
             completion(identifiers, .none)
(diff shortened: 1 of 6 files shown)


AEPIdentity/Tests/MobileIdentitiesTests.swift
```diff
@@ -30,18 +31,17 @@ class MobileIdentitiesTests: XCTestCase {
...
         let customIdTwo = CustomIdentity(origin: "origin2", type: "type2", identifier: "id2", authenticationState: .loggedOut)
         let customIdThree = CustomIdentity(origin: "origin3", type: "DSID_20915", identifier: "test-advertisingId", authenticationState: .loggedOut)
-
-        identitySharedState[IdentityConstants.EventDataKeys.VISITOR_IDS_LIST] = [customIdOne, customIdTwo, customIdThree].map({$0.asDictionary()})
-        identitySharedState[IdentityConstants.EventDataKeys.ADVERTISING_IDENTIFIER] = "test-advertisingId"
-        identitySharedState[IdentityConstants.EventDataKeys.PUSH_IDENTIFIER] = "test-pushid"
-
-        return identitySharedState
+        
+        var properties = IdentityProperties()
```
kevinlind, line 38: These tests were passing correctly before and after my changes to the public API. However, I don't feel comfortable about how this setup assumes the structure of the VISITOR_IDS_LIST. Instead, I feel it's safer to use the actual IdentityProperties to build the event data.

cdhoffmann:
Changes look good to me, but any idea why codecov is failing? Seems like it should be covered based on the tests you added? @kevinlind

cdhoffmann (approved):

kevinlind:
> Changes look good to me, but any idea why codecov is failing? Seems like it should be covered based on the tests you added? @kevinlind

It appears the project isn't configured to collect coverage on the integration tests, so the tests I added there don't count toward the coverage numbers.

AEPIdentity/Sources/Identity+PublicAPI.swift
```diff
@@ -54,11 +54,24 @@ import Foundation
...
             }
 
+            // return empty list if no Customer Identifiers in response data
+            if eventData[IdentityConstants.EventDataKeys.VISITOR_IDS_LIST] == nil {
+                completion([], nil)
+                return
+            }
+
+            guard let identifiersDict = eventData[IdentityConstants.EventDataKeys.VISITOR_IDS_LIST] as? [[String: Any]] else {
+                completion(nil, AEPError.unexpected)
```
emdobrin, line 69: can we log a debug message here in case the type cast fails?

AEPIdentity/Sources/Identity+PublicAPI.swift
```diff
@@ -143,3 +156,17 @@ import Foundation
         }
     }
 }
+
+extension CustomIdentity {
```
emdobrin, line 160: We should move this to CustomIdentity.swift

AEPIdentity/Sources/Identity+PublicAPI.swift
```diff
@@ -143,3 +156,17 @@ import Foundation
...
+    internal static func from(dict: [String: Any]) -> CustomIdentity? {
+        guard let type = dict[CodingKeys.type.rawValue] as? String,
+              let origin = dict[CodingKeys.origin.rawValue] as? String,
```
emdobrin, line 163: I think only idType and identifier values are required, but origin is optional, so we should not fail if nil. will you double check this?

AEPIdentity/Sources/Identity+PublicAPI.swift
```diff
@@ -143,3 +156,17 @@ import Foundation
...
+              let identifier = dict[CodingKeys.identifier.rawValue] as? String,
+              let rawState = dict[CodingKeys.authenticationState.rawValue] as? Int,
+              !identifier.isEmpty else {
+            return nil
```
emdobrin, line 167: I assume these ids should be already cleanup up by the internal Identity extension, but in case this fails the conversion let's log a debug message

kevinlind: Another option is to not have this check as the values are coming directly from the extension, and simply return the values given.

AEPIdentity/Sources/Identity+PublicAPI.swift
kevinlind: Origin is set internally for each of the two types of CustomIdentifiers. I thinking, however, it may be better to not check any of the values, and just take them as they are given by the extension so there is only one place which validates the data.

AEPIdentity/Sources/Identity+PublicAPI.swift
kevinlind: Yes, I wasn't sure whether to place this in CustomIdentity where others could use it, or just where it's needed in Identity+PublicAPI. I'll move it to the CustomIdentity class.

praveek (approved):

emdobrin (approved):

PR #761, opened 2022-01-13
Merging IAM changes into dev

sbenedicadb:


AEPCore.podspec
```diff
@@ -16,8 +16,8 @@ Pod::Spec.new do |s|
 
   s.pod_target_xcconfig = { 'BUILD_LIBRARY_FOR_DISTRIBUTION' => 'YES' }
 
-  s.dependency 'AEPRulesEngine', '1.0.1'
-  s.dependency 'AEPServices', '3.3.2'
+  s.dependency 'AEPRulesEngine', '1.1.0'
```
praveek, line 19: Can you change the dependency to '>= 1.1.0' similar to AEPServices?

AEPCore/Sources/eventhub/history/EventHistoryRequest.swift
```diff
@@ -0,0 +1,43 @@
...
+ Unless required by applicable law or agreed to in writing, software distributed under
+ the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ OF ANY KIND, either express or implied. See the License for the specific language
+ governing permissions and limitations under the License.
+ */
+
+import Foundation
+
+/// Used for selecting or deleting Events from Event History.
+@objc
```
praveek, line 16: It is better if we add AEP prefix for objc classes.

AEPCore/Sources/eventhub/history/EventHistoryResult.swift
```diff
@@ -0,0 +1,43 @@
...
+ the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ OF ANY KIND, either express or implied. See the License for the specific language
+ governing permissions and limitations under the License.
+ */
+
+import Foundation
+
+/// Passed to handlers by `EventHistory` when Events are requested via `getEvents` API.
+@objc
+public class EventHistoryResult: NSObject {
```
praveek, line 17: It is better if we add AEP prefix for objc classes.

AEPServices/Sources/dataqueue/SQLiteWrapper.swift
```diff
@@ -75,7 +75,7 @@ internal struct SQLiteWrapper {
     ///   - database: the database connection
     ///   - sql: the SQL statement
     /// - Returns: an `Optional` result of a database  query
-    static func query(database: OpaquePointer, sql: String) -> [[String: String]]? {
+    public static func query(database: OpaquePointer, sql: String) -> [[String: String]]? {
```
praveek, line 78: We had some concerns around this APIs when we previously discussed making this class public. Is it possible to represent the return values better instead of String?

AEPServices/Sources/ui/UIService.swift
```diff
@@ -16,21 +16,35 @@ import UIKit
 ///
 /// UIService for creating UI elements
 ///
+@objc
```
praveek, line 19: I noticed that ServiceProvider is not exposed to Objective-C. Any use case on why we need to expose this class? If so, we should prefix it with `AEP`

AEPServices/Sources/ui/fullscreen/FullscreenMessage.swift
```diff
@@ -16,23 +16,32 @@ import WebKit
 
 /// This class is used to create and display fullscreen messages on the current view.
 @objc(AEPFullscreenMessage)
-public class FullscreenMessage: NSObject, WKNavigationDelegate, FullscreenPresentable {
+public class FullscreenMessage: NSObject, FullscreenPresentable {
```
praveek, line 19: Can we move this implementation behind an interface? The direct dependency on WebKit in this class will make it difficult to port the changes over to tvOS.

AEPCore/Sources/eventhub/history/EventHistoryDatabase.swift
```diff
@@ -0,0 +1,232 @@
...
+    /// If no database connection is available, the handler will be called with a count of 0.
+    /// If there are no matching records, the handler will be called with count of 0.
+    ///
+    /// - Parameters:
+    ///   - hash: the 32-bit FNV-1a hashed representation of an Event's data.
+    ///   - from: represents the lower bounds of the date range to use when searching for the hash
+    ///   - to: represents the upper bounds of the date range to use when searching for the hash
+    ///   - handler: a callback which will contain `EventHistoryResult` representing matching events
+    func select(hash: UInt32, from: Date? = nil, to: Date? = nil, handler: @escaping (EventHistoryResult) -> Void) {
+        dispatchQueue.sync {
```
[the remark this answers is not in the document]
sbenedicadb: the `dispatchQueue` should handle any potential race conditions.  `select` is an async method, but we are accessing the `dispatchQueue` synchronously because the caller is waiting for a result.

AEPServices/Sources/dataqueue/SQLiteWrapper.swift
sbenedicadb: i'm open to suggestions as long as it doesn't break backwards compatibility with our internal code that uses the wrapper.  

did you have a specific idea on what could be returned other than an array of dictionaries (each dictionary representing a record from the database)?

AEPServices/Sources/ui/UIService.swift
sbenedicadb: `UIService` is a protocol.

i normally wouldn’t hesitate to add an `AEP` prefix to it, except we already have a class named `AEPUIService` (which implements `UIService`).

thoughts?

AEPServices/Sources/ui/fullscreen/FullscreenMessage.swift
sbenedicadb: i agree this is a good change to make.  the dependency on WebKit didn't get added to this PR, so i'd think these changes should be made when we add support for tvOS.  

thoughts?

praveek (approved):

PR #763, opened 2022-01-13
[MOB-15488] Add clearUpdatedConfiguration() API

cdhoffmann:
Adds an API which clears programmatically set config changes which have been set via the `updateConfigurationWith(configDict:)` API, as well as any update made via the `setPrivacyStatus` API. It will clear the programmatic config from cache, and revert the config to the initially set configuration. 

## New Public API:

The new public API: `clearUpdatedConfiguration()` generates a new configuration request content event called "Clear Updated Configuration" whose event data simply holds a hardcoded boolean `true` with the key: `"config.clearUpdates"`.

The Configuration Extension handles the `clearUpdatedConfiguration` event by doing the following:

1. Clearing the programmatic config from data store
2. Replacing the current configuration with the initial configuration which was set via `configureWithAppId()`, `configureWithFilePath()`, or a fallback bundled configuration
3. Resolves shared state with the cleared config
4. dispatches a configuration response content event with the cleared config
5. Notifies the rules engine about the configuration change

## Scenarios:

configureWithAppId -> updateConfigurationWithDict -> clearUpdatedConfig = The config will be the one set via configureWithAppId

configureWithFilePath -> updateConfigurationWithDict -> clearUpdatedConfig = The config will be the one set via configureWithFilePath

configureWithAppId -> updateConfigurationWithDict -> clearUpdatedConfig -> updateConfigurationWithDict = The config will be the most recently updated config, will not have any keys from the first update unless they are included in the most recent update.

configureWithAppId / configureWithFilePath -> clearUpdatedConfig = no change


AEPCore/Sources/configuration/Event+Configuration.swift
```diff
@@ -24,6 +24,11 @@ extension Event {
         return data?[ConfigurationConstants.Keys.RETRIEVE_CONFIG] as? Bool ?? false
     }
 
+    /// Returns true if this `Event` is a clear configuration event, otherwise false
+    var isClearConfigEvent: Bool {
+        return data?[ConfigurationConstants.Keys.CLEAR_CONFIG] != nil
```
sbenedicadb, line 29: may be better to check the value here, rather than just the existence of the property.  similar to `isGetConfigEvent` above?

sbenedicadb (approved):
looks good.  take or leave the change i suggested.

AEPCore/Sources/configuration/ConfigurationConstants.swift
```diff
@@ -28,6 +28,7 @@ struct ConfigurationConstants {
         static let GLOBAL_CONFIG_PRIVACY = "global.privacy"
         static let UPDATE_CONFIG = "config.update"
         static let RETRIEVE_CONFIG = "config.getData"
+        static let CLEAR_CONFIG = "config.clear"
```
praveek, line 31: Let us change this value to config.clearUpdates.

praveek (approved):

PR #782, opened 2022-03-08
[MOB-15968] Implemented logic to send ecid and customIds without waiting for config.

addb:
## Types of changes

## Checklist:

### DIFF
--- a/AEPIdentity/Sources/Identity.swift
+++ b/AEPIdentity/Sources/Identity.swift
@@ -58,16 +58,15 @@ import Foundation
         guard let state = state else { return false }
 
         // fast boot identity without waiting for configuration
-        if state.boot(event: event, createSharedState: createSharedState(data:event:)) {
-
-            // skip waiting for configuration if event is getExperienceCloudId event or getIdentifiers event
-            if event.isGetIdentifierEvent {
-                Log.trace(label: "\(name):\(#function)", "Not waiting for latest configuration as we have the ECID cached and available for processing [event:(\(event.name)) id:(\(event.id)].")
-                return true
-            }
-        }
+        state.boot(event: event, createSharedState: createSharedState(data:event:))
 
         guard state.forceSyncIdentifiers(configSharedState: getSharedState(extensionName: IdentityConstants.SharedStateKeys.CONFIGURATION, event: nil)?.value, event: event, createSharedState: createSharedState(data:event:)) else { return false }
+
+        // skip waiting for latest configuration if it is getExperienceCloudId event or getIdentifiers event
+        if event.isGetIdentifierEvent {
+            Log.trace(label: "\(name):\(#function)", "Not waiting for latest configuration as we have the cached ECID available for processing [event:(\(event.name)) id:(\(event.id)].")
+            return true
+        }
 
         if event.isSyncEvent || event.type == EventType.genericIdentity {
             guard let configSharedState = getSharedState(extensionName: IdentityConstants.SharedStateKeys.CONFIGURATION, event: event)?.value else {
@@ -76,13 +75,13 @@ import Foundation
             }
             return state.readyForSyncIdentifiers(event: event, configurationSharedState: configSharedState)
         } else if event.type == EventType.configuration, event.source == EventSource.requestIdentity {
-            let areSharedStateReady = MobileIdentities().areSharedStatesReady(event: event, sharedStateProvider: getSharedState(extensionName:event:))
+            let areSharedStatesReady = MobileIdentities().areSharedStatesReady(event: event, sharedStateProvider: getSharedState(extensionName:event:))
 
-            if !areSharedStateReady {
+            if !areSharedStatesReady {
                 Log.trace(label: "\(name):\(#function)", "Waiting for the Mobile Identities states to be set before processing [event:(\(event.name)) id:(\(event.id)].")
             }
 
-            return areSharedStateReady
+            return areSharedStatesReady
         } else if event.type == EventType.identity, event.source == EventSource.requestIdentity, ( event.baseUrl != nil ||  event.urlVariables ) {
 
             // analytics shared state will be null if analytics extension is not registered. Wait for analytics shared only if the status is pending or none
--- a/AEPIdentity/Sources/Identity.swift
+++ b/AEPIdentity/Sources/Identity.swift
@@ -66,9 +66,7 @@ import Foundation
         if event.isGetIdentifierEvent {
             Log.trace(label: "\(name):\(#function)", "Processing get identifier event without waiting for configuration [event:(\(event.name)) id:(\(event.id)].")
             return true
-        }
-
-        if event.isSyncEvent || event.type == EventType.genericIdentity {
+        } else if event.isSyncEvent || event.type == EventType.genericIdentity {
             guard let configSharedState = getSharedState(extensionName: IdentityConstants.SharedStateKeys.CONFIGURATION, event: event)?.value else {
                 Log.trace(label: "\(name):\(#function)", "Waiting for the Configuration shared state value before processing [event:(\(event.name)) id:(\(event.id)].")
                 return false
--- a/AEPIdentity/Sources/Identity.swift
+++ b/AEPIdentity/Sources/Identity.swift
@@ -55,23 +55,31 @@ import Foundation
     }
 
     public func readyForEvent(_ event: Event) -> Bool {
-        // fast boot identity without waiting for configuration
-        guard bootUp(event: event) else { return false }
+        guard let state = state else { return false }
 
-        // skip waiting for configuration if event is getExperienceCloudId event or getIdentifiers event
-        if isGetIdentifierEvent(event: event) {
-            Log.trace(label: "\(name):\(#function)", "Not waiting for latest configuration as we have the ECID cached and available for processing [event:(\(event.name)) id:(\(event.id)].")
-            return true
+        // fast boot identity without waiting for configuration
+        if state.boot(event: event, createSharedState: createSharedState(data:event:)) {
+
+            // skip waiting for configuration if event is getExperienceCloudId event or getIdentifiers event
+            if event.isGetIdentifierEvent {
+                Log.trace(label: "\(name):\(#function)", "Not waiting for latest configuration as we have the ECID cached and available for processing [event:(\(event.name)) id:(\(event.id)].")
+                return true
+            }
         }
 
-        guard canProcessEvents(event: event) else { return false }
+        guard let configurationSharedState = getSharedState(extensionName: IdentityConstants.SharedStateKeys.CONFIGURATION, event: nil)?.value else {
+            Log.trace(label: "\(name):\(#function)", "Waiting for the Configuration shared state to get required configuration fields before processing [event:(\(event.name)) id:(\(event.id)].")
+            return false
+        }
+
+        guard state.forceSyncIdentifiers(configSharedState: configurationSharedState, event: event) else { return false }
 
         if event.isSyncEvent || event.type == EventType.genericIdentity {
             guard let configSharedState = getSharedState(extensionName: IdentityConstants.SharedStateKeys.CONFIGURATION, event: event)?.value else {
-                Log.trace(label: "\(name):\(#function)", "Waiting for the Analytics shared state value before processing event: (\(event.name)).")
+                Log.trace(label: "\(name):\(#function)", "Waiting for the Configuration shared state value before processing [event:(\(event.name)) id:(\(event.id)].")
                 return false
             }
-            return state?.readyForSyncIdentifiers(event: event, configurationSharedState: configSharedState) ?? false
+            return state.readyForSyncIdentifiers(event: event, configurationSharedState: configSharedState)
         } else if event.type == EventType.configuration, event.source == EventSource.requestIdentity {
             let areSharedStateReady = MobileIdentities().areSharedStatesReady(event: event, sharedStateProvider: getSharedState(extensionName:event:))
 
@@ -96,53 +104,6 @@ import Foundation
         }
 
         return isConfigSharedStateSet
-    }
-
-    // Loads all the identifiers in identity state from persistence, generates ecid if not found in persistence and gets the extension ready for get identifiers event.
-    /// - Parameter event: An `Event`
-    /// - Returns: True if we can process events, false otherwise
-    private func bootUp(event: Event) -> Bool {
-        guard let state = state else { return false }
-        guard !state.hasBooted else { return true } // we have booted, return true
-
-        // attempt to bootup
-        if state.boot(event: event) {
-            createSharedState(data: state.identityProperties.toEventData(), event: nil)
-            return true
-        }
-
-        return false // cannot handle any events until we have booted
-
-    }
-
-    /// Determines if Identity is ready to handle events, this is determined by if the Identity extension has received all the required configuration
-    /// - Parameter event: An `Event`
-    /// - Returns: True if we can process events, false otherwise
-    private func canProcessEvents(event: Event) -> Bool {
-        guard let state = state else { return false }
-        guard !state.hasSynced else { return true } // we have synced, return true
-
-        // fetch latest configuration when booting up
-        guard let configSharedState = getSharedState(extensionName: IdentityConstants.SharedStateKeys.CONFIGURATION, event: nil)?.value else {
-            Log.trace(label: "\(name):\(#function)", "Waiting for the Configuration shared state to get required configuration fields before processing [event:(\(event.name)) id:(\(event.id)].")
-            return false
-        }
-
-        if state.SyncIdentifiersIfConfigurationAvailable(configSharedState: configSharedState, event: event) {
-            return true
-        }
-
-        Log.trace(label: "\(name):\(#function)", "Waiting for the Identity state to Sync up. Identity will sync up when it has recevied all the required configuration before processing [event:(\(event.name)) id:(\(event.id)].")
-
-        return false // cannot handle any events until we have booted
-    }
-
-    /// Determines if Identity is event is getExperienceCloudId event or getIdentifiers event
-    /// - Parameter event: An `Event`
-    /// - Returns: true if the event is getExperienceCloudId event or getIdentifiers event
-    private func isGetIdentifierEvent(event: Event) -> Bool {
-        // return true if the event is getExperienceCloudId event or getIdentifiers event
-        return event.type == EventType.identity && event.source == EventSource.requestIdentity && event.baseUrl == nil && !event.urlVariables
     }
 
     // MARK: Event Listeners
--- a/AEPIdentity/Sources/Identity.swift
+++ b/AEPIdentity/Sources/Identity.swift
@@ -55,23 +55,45 @@ import Foundation
     }
 
     public func readyForEvent(_ event: Event) -> Bool {
+        // check if getExperienceCloudId or getIdentifiers event and if ecid is cached.
+        // if both the conditions are met allow getExperienceCloudId or getIdentifiers events to continue without waiting for configuration shared state.
+        if isGetIdentifierEvent(event: event) && isECIDCached() {
+            Log.trace(label: "\(name):\(#function)", "Bypassing canProcessEvents() and not waiting for latest configuration as we have the ECID cached and available for processing event: (\(event.name)).")
+            return true
+        }
+
         guard canProcessEvents(event: event) else { return false }
 
         if event.isSyncEvent || event.type == EventType.genericIdentity {
-            guard let configSharedState = getSharedState(extensionName: IdentityConstants.SharedStateKeys.CONFIGURATION, event: event)?.value else { return false }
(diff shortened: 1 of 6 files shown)


emdobrin (commented):
@addb I am thinking about two changes required here:
- the getters should always fetch current identities and not wait for configuration
- initial shared state should be shared on boot if there is an existing identity

addb:
> @addb I am thinking about two changes required here:
> 
> * the getters should always fetch current identities and not wait for configuration
> * initial shared state should be shared on boot if there is an existing identity

When you say share on boot, does theat mean the logic where we check for state.hasBooted flag or we want to share state if we see cached identifier value even if we dont have required config(i.e state.hasBooted = false)?

kevinlind:
> @addb I am thinking about two changes required here:
> 
> * the getters should always fetch current identities and not wait for configuration
> * initial shared state should be shared on boot if there is an existing identity

@emdobrin Can you elaborate on the second point? From my understanding, on first launch Identity waits for configuration to load before generating the ECID and shared state, same behavior as before (let me know if that requirement changed, but I remember that being the decision). On subsequent launches, however, the cached configuration will load and allow Identity to boot before the remote configuration is fetched again. Unless there's a case I'm missing, I see the current implementation as correctly setting the shared state on boot.

Another concern I have is this implementation of ignoring any pending shared states to fetch the identities is fine if the event queue is empty, however if there is another event waiting on configuration to resolve (such as getUrlVariables()), then the getExperienceCloudId() call will still wait in the queue. Do we need to handle such a case?

AEPIdentity/Sources/Identity.swift
```diff
@@ -55,41 +55,94 @@ import Foundation
...
+        }
+
+        if state.SyncIdentifiersIfConfigurationAvailable(configSharedState: configSharedState, event: event) {
+            return true
         }
 
+        Log.trace(label: "\(name):\(#function)", "Waiting for the Identity state to Sync up. Identity will sync up when it has recevied all the required configuration before processing [event:(\(event.name)) id:(\(event.id)].")
+
         return false // cannot handle any events until we have booted
+    }
```
emdobrin, line 138: return state.SyncIdentifiersIfConfigurationAvailable(configSharedState: configSharedState, event: event)

AEPIdentity/Sources/IdentityState.swift
```diff
@@ -41,28 +42,38 @@ class IdentityState {
...
+        // generate ECID is not found in persistence
+        generateAndPersistECID()
+
+        hasBooted = true
+        Log.debug(label: "\(LOG_TAG):\(#function)", "Identity has successfully booted up")
+
+        return hasBooted
+    }
+
+    func SyncIdentifiersIfConfigurationAvailable(configSharedState: [String: Any], event: Event) -> Bool {
```
emdobrin, line 59: renaming suggestion: forceSyncIdentifiers

Also same comment here:
if hasSynced {
return true
}

AEPIdentity/Sources/IdentityState.swift
```diff
@@ -41,28 +42,38 @@ class IdentityState {
     ///   - configSharedState: the current configuration shared state available at registration time
     ///   - event: The `Event` triggering the bootup
     /// - Returns: True if we should share state after bootup, false otherwise
-    func bootupIfReady(configSharedState: [String: Any], event: Event) -> Bool {
+    func boot(event: Event) -> Bool {
```
emdobrin, line 45: I think we should keep the control for hasBooted inside of this function and with the change below I think we can remove the bootUp func, 
e.g.
if hasBooted {
return true
}

AEPIdentity/Sources/IdentityState.swift
```diff
@@ -19,6 +19,7 @@ class IdentityState {
     private(set) var hitQueue: HitQueuing
     private var pushIdManager: PushIDManageable
     private(set) var hasBooted = false
+    private(set) var hasSynced = false
```
emdobrin, line 22: I think we can now make these private set/get

AEPIdentity/Sources/Identity.swift
addb: I wanted to have log for false case.

AEPIdentity/Sources/Identity.swift
```diff
@@ -55,41 +55,55 @@ import Foundation
...
+            // skip waiting for configuration if event is getExperienceCloudId event or getIdentifiers event
+            if event.isGetIdentifierEvent {
+                Log.trace(label: "\(name):\(#function)", "Not waiting for latest configuration as we have the ECID cached and available for processing [event:(\(event.name)) id:(\(event.id)].")
+                return true
+            }
+        }
+
+        guard let configurationSharedState = getSharedState(extensionName: IdentityConstants.SharedStateKeys.CONFIGURATION, event: nil)?.value else {
+            Log.trace(label: "\(name):\(#function)", "Waiting for the Configuration shared state to get required configuration fields before processing [event:(\(event.name)) id:(\(event.id)].")
+            return false
```
kevinlind, line 72: Here you are getting the latest Configuration state for each Identity event, but it is only needed when booting and performing the first force sync event. You could wrap this in an `if !state.hasSynced ...` block, but I think it would be cleaner to pass the `getSharedState` function to `IdentityState.forceSyncIdentifiers` so all the logic to validate the initial force sync event is handled there.

AEPIdentity/Sources/Identity.swift
```diff
@@ -55,41 +55,55 @@ import Foundation
...
+                Log.trace(label: "\(name):\(#function)", "Waiting for the Configuration shared state value before processing [event:(\(event.name)) id:(\(event.id)].")
+                return false
+            }
+            return state.readyForSyncIdentifiers(event: event, configurationSharedState: configSharedState)
         } else if event.type == EventType.configuration, event.source == EventSource.requestIdentity {
-            return MobileIdentities().areSharedStatesReady(event: event, sharedStateProvider: getSharedState(extensionName:event:))
+            let areSharedStateReady = MobileIdentities().areSharedStatesReady(event: event, sharedStateProvider: getSharedState(extensionName:event:))
+
+            if !areSharedStateReady {
+                Log.trace(label: "\(name):\(#function)", "Waiting for the Analytics, Identity, Configuration and Audience shared states to be set before processing [event:(\(event.name)) id:(\(event.id)].")
```
kevinlind, line 87: Listing the extension names can become in correct if MobileIdentities is updated (see the TODO to add Target?). Instead, rename to something like "Waiting for Mobile Identities states to be set before processing ...".

AEPIdentity/Sources/IdentityState.swift
```diff
@@ -36,33 +37,51 @@ class IdentityState {
...
         // Update hit queue with privacy status
         hitQueue.handlePrivacyChange(status: identityProperties.privacyStatus)
 
-        hasBooted = true
-        Log.debug(label: "\(LOG_TAG):\(#function)", "Identity has successfully booted up")
         // Identity should always share its state
         // However, don't create a shared state twice, which will log an error
-        // The force sync event processed above will create a shared state if the privacy is not opt-out
-        return syncIdentifiers(event: event) != nil || identityProperties.privacyStatus == .optedOut
+        hasSynced = syncIdentifiers(event: event) != nil || identityProperties.privacyStatus == .optedOut
```
kevinlind, line 82: If `hasSynced` is true, but `hasBooted` is false, then we need to create a shared state with the current IdentityProperties, event if there are none. It is important that Identity create a shared state after handling the boot event, event if that state is empty.
After creating the shared state, set `hasBooted` to true.

kevinlind (changes requested):

kevinlind:
Can you also add some unit tests to cover the lines Codecov is flagging, to increase the test coverage of the changes you've made?

AEPIdentity/Sources/Identity.swift
```diff
@@ -55,41 +55,50 @@ import Foundation
...
+        if state.boot(event: event, createSharedState: createSharedState(data:event:)) {
+
+            // skip waiting for configuration if event is getExperienceCloudId event or getIdentifiers event
+            if event.isGetIdentifierEvent {
+                Log.trace(label: "\(name):\(#function)", "Not waiting for latest configuration as we have the ECID cached and available for processing [event:(\(event.name)) id:(\(event.id)].")
+                return true
+            }
+        }
+
+        guard state.forceSyncIdentifiers(configSharedState: getSharedState(extensionName: IdentityConstants.SharedStateKeys.CONFIGURATION, event: nil)?.value, event: event, createSharedState: createSharedState(data:event:)) else { return false }
```
emdobrin, line 70: As we discussed since the extension does not have access to the events queue, we should not do anything to change the order of the events here.
The only change should be when a getIdentifierEvent is received we don't check for config. The sequence would look like:
```
state.boot(event: event, createSharedState: createSharedState(data:event:))
if !state.forceSyncIdentifiers(configSharedState: getSharedState(extensionName: IdentityConstants.SharedStateKeys.CONFIGURATION, event: nil)?.value, event: event, createSharedState: createSharedState(data:event:)) else { return false }
if event.isGetIdentifierEvent { return true } // does not wait for config
if event.isSyncEvent ...
```

This should also account for the case where ECID is not generated yet on first launch, which will be generated on the forceSyncIdentifiers step.

AEPIdentity/Sources/Identity.swift
```diff
@@ -55,41 +55,50 @@ import Foundation
...
-            guard let configSharedState = getSharedState(extensionName: IdentityConstants.SharedStateKeys.CONFIGURATION, event: event)?.value else { return false }
-            return state?.readyForSyncIdentifiers(event: event, configurationSharedState: configSharedState) ?? false
+            guard let configSharedState = getSharedState(extensionName: IdentityConstants.SharedStateKeys.CONFIGURATION, event: event)?.value else {
+                Log.trace(label: "\(name):\(#function)", "Waiting for the Configuration shared state value before processing [event:(\(event.name)) id:(\(event.id)].")
+                return false
+            }
+            return state.readyForSyncIdentifiers(event: event, configurationSharedState: configSharedState)
         } else if event.type == EventType.configuration, event.source == EventSource.requestIdentity {
-            return MobileIdentities().areSharedStatesReady(event: event, sharedStateProvider: getSharedState(extensionName:event:))
+            let areSharedStateReady = MobileIdentities().areSharedStatesReady(event: event, sharedStateProvider: getSharedState(extensionName:event:))
```
emdobrin, line 79: typo: areSharedStatesReady

AEPIdentity/Sources/Identity.swift
```diff
@@ -55,41 +55,50 @@ import Foundation
...
-        // attempt to bootup
-        if state.bootupIfReady(configSharedState: configSharedState, event: event) {
-            createSharedState(data: state.identityProperties.toEventData(), event: nil)
+        if !isConfigSharedStateSet {
+            Log.trace(label: "\(name):\(#function)", "Waiting for the Configuration shared state to be set before processing [event:(\(event.name)) id:(\(event.id)].")
         }
 
-        return false // cannot handle any events until we have booted
-
+        return isConfigSharedStateSet
```
emdobrin, line 101: What other events are handled by this case? Does resetIdentities for example also wait for config?

AEPIdentity/Sources/IdentityState.swift
```diff
@@ -57,12 +84,17 @@ class IdentityState {
...
+
+        // If the sync was susccessful and there is no shared state update, post a shared state update
+        if hasSynced && !hasBooted {
```
emdobrin, line 92: See my comment above, this would become: 
```
if hasSynced && !didCreateFirstSharedState {
     createSharedState(identityProperties.toEventData(), event)
     didCreateBootSharedState = true
}

File: AEPIdentity/Sources/IdentityState.swift (hunk 7 above)
Comment (emdobrin) [reply AEPIdentity/Sources/IdentityState.swift:89]: Also, here `syncIdentifiers` should not create a shared state if it was already created on boot.

File: AEPIdentity/Sources/IdentityState.swift [hunk 12]
```diff
@@ -205,11 +237,10 @@ class IdentityState {
         if identityProperties.ecid != nil, !hasIds, !hasDpids, !needResync {
             Log.trace(label: "\(LOG_TAG):\(#function)", "Not syncing identifiers at this time, no new identifiers or previously synced.")
             syncForIds = false
-        } else if identityProperties.ecid == nil {
-            Log.trace(label: "\(LOG_TAG):\(#function)", "Generating new ECID, ECID not found in persistence.")
-            identityProperties.ecid = ECID()
         }
 
+        generateAndPersistECID()
```
emdobrin, line 242: there might be a slight change in logic here since this is no longer on the else branch. Should we keep 
``` 
if identityProperties.ecid != nil, !hasIds, !hasDpids, !needResync {
... 
} else {
   generateAndPersistECID()
}

File: AEPIdentity/Sources/Identity.swift (hunk 8 above)
Comment (addb) [reply AEPIdentity/Sources/Identity.swift:70]: forceSyncIdentifiers requires valid config and if we have pending shared state then this will again not work right? Is the change for not setting pending shared state for configuration going to happen?

File: AEPIdentity/Sources/IdentityState.swift (hunk 12 above)
Comment (addb) [reply AEPIdentity/Sources/IdentityState.swift:242]: I was thinking it will generate ecid if there was none before and doesnt regenerate if already present. I was trying to simplify the logic, since if ecid is nil it is going to anyways fall to this code.

File: AEPIdentity/Sources/Identity.swift (hunk 8 above)
Comment (emdobrin) [reply AEPIdentity/Sources/Identity.swift:70]: @addb we should keep the two changes separately: this one is specifically to return readyForEvent without waiting for config and the not setting pending shared state changes will solve for having the config resolved quickly, so even if the getIdentifier event is queued after a force sync event it will return pretty quickly.

Comment (kevinlind) [reply AEPIdentity/Sources/Identity.swift:70]: @emdobrin Even without setting a pending state, when the SDK starts, if an app has no bundled config and only configures with app id, then readyForEvent will still return false until that first configuration request is resolved. From your recommendation, the shared state is created with the persisted ECID immediately, but any getIdentifier events will wait for the initial configuration. I'm not clear on that separation and why we don't handle both at the same time, either both after `state.boot` or after the force sync event.

File: AEPIdentity/Sources/Identity.swift (hunk 10 above)
Comment (addb) [reply AEPIdentity/Sources/Identity.swift:100]: @emdobrin we wait for config for almost all events at present, except for have isSyncEvent or genericIdentity event type where we look for readyForSyncIdentifiers check

Comment (emdobrin) [reply AEPIdentity/Sources/Identity.swift:100]: Ok, we may need to evaluate the other events also, e.g. resetIdentities should not have a dependency on config, but I think we can keep that outside of this PR.

File: AEPIdentity/Tests/IdentityStateTests.swift [hunk 13]
```diff
@@ -33,63 +33,159 @@ class IdentityStateTests: XCTestCase {
...
+    func testBootECIDUnavailable() {
         // test
-        let result = state.bootupIfReady(configSharedState: [:], event: Event.fakeSyncIDEvent())
+        let result = state.boot(event: Event.fakeSyncIDEvent(), createSharedState: { (data, event) in
+            XCTFail("Shared state should not be updated")
+        })
+    }
+
+    /// Tests that the there is a shared state update when ecid is present and not wait for configuration shared state
+    func testBootECIDAvailableWithoutConfiguration() {
```
emdobrin, line 47: testBoot_whenECID_createsInitialSharedState

AEPIdentity/Tests/IdentityStateTests.swift
```diff
@@ -33,63 +33,159 @@ class IdentityStateTests: XCTestCase {
...
+        })
+
+        // verify
+        wait(for: [sharedStateExpectation], timeout: 1)
+    }
+
+    // MARK: forceSyncIdentifiers(...) tests
+
+    /// Tests that force sync fails and there is no shared state created when empty configuration shared state is received
+    func testForceSyncIdentifiersEmptyConfigSharedState() {
```
emdobrin, line 66: testForceSyncIdentifiers_whenEmptyConfig_returnsFalse_noInitialSharedState

emdobrin (commented):
The changes look good, the remaining comments are mainly on logs and tests.

AEPIdentity/Sources/Identity.swift
emdobrin: Sharing the initial shared state on boot is still required for other extensions or data elements to consume if it already has identities, so that change by itself will improve certain workflows.
With regards to not retuning on getters right away on boot, I do not see a reason why we should speed up the process for get identities events only if they are received first by this extension (on boot), but not in other use-cases, that could become difficult to debug and maintain over time. 
However, I 💯  agree we should revisit this point if consuming the cached configuration and not block on pending state does not solve for this use-case.

AEPIdentity/Tests/IdentityFunctionalTests.swift
```diff
@@ -383,18 +388,16 @@ class IdentityFunctionalTests: XCTestCase {
         XCTAssertEqual(EventSource.responseIdentity, dispatchedEvent?.source)
         XCTAssertNotNil(dispatchedEvent?.data?[IdentityConstants.EventDataKeys.VISITOR_ID_ECID])
         let ecid = dispatchedEvent?.data?[IdentityConstants.EventDataKeys.VISITOR_ID_ECID] as? String ?? ""
-        XCTAssertFalse(ecid.isEmpty);
+        XCTAssertFalse(ecid.isEmpty)
     }
 
-    /// Tests that getExperienceCloudId doesnt wait for Identity bootup and ECID to be generated when cached ECID is not avaiable
-    func testGetECIDWaitsForConfigurationSharedStateToResolveWhenCachedECIDNotAvailable() {
+    func testGetECIDAfterForceSyncDoesNotWaitForConfigurationSharedState() {
```
emdobrin, line 394: for this test we need to set the shared state on pending after force sync:
1. set valid config
2. dispatch any event -> initiates force sync
3. set config nil data and .pending
4. dispatch get ecid event
5. assert returns data

AEPIdentity/Tests/IdentityTests.swift
```diff
@@ -548,7 +545,25 @@ class IdentityTests: XCTestCase {
         mockRuntime.simulateSharedState(extensionName: "com.adobe.module.analytics", event: getUrlVariablesEvent, data: ([IdentityConstants.Analytics.ANALYTICS_ID: "test-aid"], .set))
 
         // verify
-        XCTAssertFalse(identity.readyForEvent(getUrlVariablesEvent)) // booting up
-        XCTAssertTrue(identity.readyForEvent(getUrlVariablesEvent))
+        XCTAssertTrue(identity.readyForEvent(getUrlVariablesEvent)) // fast boot
+    }
+
+    func testReadyForEventGetEcidOnAppInstallWillWaitForConfigurationAndECIDToBeGenerated() {
```
kevinlind, line 551: Can you verify in some way that the event actually is waiting on the configuration state? Maybe something like:

1. call readyForEvent and verify False
2. verify ECID is not set in Identity Properties
3. set valid Configuration
4. call readyForEvent and verify True
5. verify ECID is set in Identity Properties.

AEPIdentity/Tests/IdentityTests.swift
```diff
@@ -548,7 +545,25 @@ class IdentityTests: XCTestCase {
...
+        let getEcidEvent = Event(name: "Test Get ECID Event", type: EventType.identity, source: EventSource.requestIdentity, data: nil)
+
+        mockRuntime.simulateSharedState(extensionName: IdentityConstants.SharedStateKeys.CONFIGURATION, event: getEcidEvent, data: ([IdentityConstants.Configuration.EXPERIENCE_CLOUD_ORGID: "test-org-id", IdentityConstants.Configuration.GLOBAL_CONFIG_PRIVACY: PrivacyStatus.optedIn.rawValue], .set))
+
+        // verify
+        XCTAssertTrue(identity.readyForEvent(getEcidEvent)) // fast boot
+    }
+
+    func testReadyForEventGetEcidOnWillWaitForConfigurationWhenEcidIsCached() {
```
kevinlind, line 560: Similar to the above test, can you verify that the event is actually waiting on the configuration state? Maybe something like:

1. set ECID in mock datastore
2. call readyForEvent and verify False
3. verify ECID is set in Identity Properties
4. set valid Configuration
5. call readyForEvent and verify True

AEPIdentity/Tests/IdentityStateTests.swift
```diff
@@ -33,65 +33,161 @@ class IdentityStateTests: XCTestCase {
...
+        let result = state.boot(event: Event.fakeSyncIDEvent(), createSharedState: { (data, event) in
+            XCTFail("Shared state should not be updated")
+        })
+    }
+
+    /// Tests that the there is a shared state update when ecid is present and not wait for configuration shared state
+    func testBoot_whenECID_createsInitialSharedState() {
+        // setup
+        let ecid = ECID()
+        state.identityProperties.ecid = ecid
```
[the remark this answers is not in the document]
addb: I will add this test in the other ticket. There is some issue with MockDataStore.

kevinlind (approved):

emdobrin (approved):

PR #783, opened 2022-03-15
[MOB-15892] [MOB-15895] Use fractional seconds in timestamps sent to Edge Network

kevinlind:
## Description
Use fractional seconds in timestamps sent to Edge Network.

- Added function to Date+Format to format timestamp with fractional seconds and UTC timezone.
- Added function to Date+Format to format timestamp to full date. Function is a port from the Edge extension in effort to consolidate date and time utility functions to Date+Format.
- Change XDMMobileLifecycleDetails to set `xdm.timestamp` using new Date+Format function to format timestamp with fractional seconds.

## Types of changes

## Checklist:


AEPCore/Tests/RulesTests/TokenFinderTests.swift
```diff
@@ -123,9 +124,9 @@ class TokenFinderTests: XCTestCase {
             return
         }
         let date_UNIX = Date(timeIntervalSince1970: TimeInterval(date_UNIX_Int))
-        /// Then: return same timestamp with different format
-        XCTAssertEqual(date_ISO8601, date_ISO8601NoColon)
-        XCTAssertEqual(date_ISO8601, date_UNIX)
+        // Then: return same timestamp with different format, but only compare seconds as ISO8601NoColon and UNIX do not count milliseconds
+        XCTAssertEqual(0, Calendar.current.dateComponents([.second], from: date_ISO8601, to: date_ISO8601NoColon).second)
+        XCTAssertEqual(0, Calendar.current.dateComponents([.second], from: date_ISO8601, to: date_UNIX).second)
```
kevinlind, line 129: As `date_ISO8601` contains milliseconds, it cannot be directly compared to the dates with only seconds.

AEPLifecycle/Tests/LifecycleSessionTests.swift
```diff
@@ -170,7 +170,7 @@ class LifecycleSessionTests: XCTestCase {
         // test
         session.start(date: currentDate, sessionTimeout: sessionTimeoutInSeconds, coreMetrics: LifecycleMetrics())
 
-        let previousSessionInfo = LifecycleSessionInfo(startDate: Calendar.current.date(byAdding: .day, value: -8, to: currentDate),
+        let previousSessionInfo = LifecycleSessionInfo(startDate: currentDate.addingTimeInterval(-8*24*60*60), // currentDate - 8 days
```
kevinlind, line 173: Using `Calendar` here will correct for daylight savings time changes. So, when the startDate is in standard time while the pauseDate is in daylight savings, the session time is "correctly" off by 3600 seconds. Better to just add/subtract from the time interval so the expected result is the same when the current time just past a daylight savings change.

AEPServices/Sources/utility/Date+Format.swift
```diff
@@ -12,23 +12,52 @@
...
+    }
+
+    /// Returns a string representation of this Date formatted as an ISO 8601 date without time, using system local time zone.
+    /// For example, Oct. 28 2020 at 9:08:32.301 am PST is returned as `2020-10-28`.
+    /// - Returns: a string representation of this Date formatted as an ISO 8601 date without time, using UTC time zone.
+    func getISO8601FullDate() -> String {
+        let formatter = ISO8601DateFormatter()
+        formatter.timeZone = TimeZone.current
+        formatter.formatOptions = [.withFullDate, .withDashSeparatorInDate]
+        return formatter.string(from: self)
```
kevinlind, line 61: I'm porting this from the Edge extension from here: https://github.com/adobe/aepsdk-edge-ios/blob/dev/Sources/xdm/XDMFormatters.swift#L54
We will eventually remove the XDMFormatters class from Edge in favor of using this public Date+Format utility class.
As the original method in Edge uses the system local timezone, I use the current timezone here instead of UTC.

AEPServices/Sources/utility/Date+Format.swift
```diff
@@ -12,23 +12,52 @@
...
         formatter.locale = Locale(identifier: "en_US_POSIX")
         formatter.dateFormat = "yyyy-MM-dd'T'HH:mm:ssXX"
         return formatter.string(from: self)
     }
+
+    /// Returns a string representation of this Date formatted as an ISO 8601 date-time with fractional seconds, using UTC time zone.
+    /// Use this date format for timestamps send to the Adobe Experience Edge Network.
+    /// For example, Oct. 28 2020 at 9:08:32.301 am PST is returned as `2020-10-28T17:08:32.301Z`
+    /// - Returns: a string representation of the given Date formatted as an ISO 8601 date-time with fractional seconds using UTC time zone.
+    func getISO8601DateInMillisecondsUTC() -> String {
```
emdobrin, line 46: renaming suggestion: getISO8601UTCDateInMilliseconds()

In this doc should we also include link to the RFC for ISO 8601?

AEPLifecycle/Tests/LifecycleV2MetricsBuilderTests.swift
```diff
@@ -157,15 +157,16 @@ class LifecycleV2MetricsBuilderTests: XCTestCase {
...
         
         XCTAssertTrue(NSDictionary(dictionary: actualAppCloseData ?? [:]).isEqual(to: expected))
     }
     
     func testBuildAppCloseXDMDataReturnsCorrectDataWhenIsCloseCorrectSession() {
         let actualAppCloseData = xdmMetricsBuilder.buildAppCloseXDMData(
-            launchDate: Date(timeIntervalSince1970: 1483864368), // start: Sunday, January 8, 2017 8:32:48 AM GMT
-            closeDate: Date(timeIntervalSince1970: 1483864390), fallbackCloseDate: Date(timeIntervalSince1970: 1483864390), // close: Sunday, January 8, 2017 8:33:10 AM GMT
+            launchDate: Date(milliseconds: 1483864368005), // start: Sunday, January 8, 2017 8:32:48 AM GMT
+            closeDate: Date(milliseconds: 1483864390123),
```
emdobrin, line 168: nit: can you print the date for this one also?

AEPServices/Sources/utility/Date+Format.swift
```diff
@@ -12,23 +12,52 @@
 import Foundation
 
 public extension Date {
+
+    /// Returns the number of seconds since the Unix Epoch on 1 January 1970..
```
emdobrin, line 16: double period at the end

AEPServices/Sources/utility/Date+Format.swift
kevinlind: I've updated the method name.  Regarding the link, I know we did that in Edge, but RFC 3339 is considered a profile of ISO 8601 and they are not 100% equivalent. Apple's docs for ISO8601DateFormatter links to the ISO site, but the specs are not available unless purchased. Java's DateTimeFormatter docs mention ISO formats but don't link to ISO or RFC.

AEPServices/Sources/utility/Date+Format.swift
```diff
@@ -12,23 +12,52 @@
...
         formatter.locale = Locale(identifier: "en_US_POSIX")
         formatter.dateFormat = "yyyy-MM-dd'T'HH:mm:ssXX"
         return formatter.string(from: self)
     }
+
+    /// Returns a string representation of this Date formatted as an ISO 8601 date-time with fractional seconds, using UTC time zone.
+    /// Use this date format for timestamps send to the Adobe Experience Edge Network.
+    /// For example, Oct. 28 2020 at 9:08:32.301 am PST is returned as `2020-10-28T17:08:32.301Z`
+    /// - Returns: a string representation of the given Date formatted as an ISO 8601 date-time with fractional seconds using UTC time zone.
+    func getISO8601UTCDateInMilliseconds() -> String {
```
praveek, line 46: nit: I feel `getISO8601UTCDateWithMilliseconds` seems more apt for this function.

AEPCore/Sources/rules/TokenFinder.swift
```diff
@@ -68,7 +68,7 @@ class TokenFinder: Traversable {
         case TOKEN_KEY_TIMESTAMP_ISO8601_NO_COLON:
             return now.getISO8601DateNoColon()
         case TOKEN_KEY_TIMESTAMP_ISO8601:
```
praveek, line 70: nit: Can you rename this constant to more appropriate name?

kevinlind: I renamed it to `TOKEN_KEY_TIMESTAMP_ISO8601_UTC_MILLISECONDS`

AEPServices/Sources/utility/Date+Format.swift
emdobrin: Thanks for the additional info, I think the standard mention and the examples you have should be sufficient.

emdobrin (approved):
Looks good. 
@praveek since these are public APIs let's do a final review with the team prior to release.

PR #787, opened 2022-03-28
[MOB-16288]SqliteDataQueue single quote fix #784

### LINKED ISSUE #784: SQLiteDataQueue fails to persist payload containing single quotes
### Expected Behaviour

DataEntity should always get persisted in SQLiteDataQueue

### Actual Behaviour

DataEntity with single quotes in payload does not get persisted in SQLiteDataQueue

### Reproduce Scenario (including but not limited to)
```
private struct EventEntity: Codable {
        var id: UUID
        var timestamp: Date
        var name: String
    }
let queue = DataQueueService().getDataQueue(label: fileName)!
let event = EventEntity(id: UUID(), timestamp: Date(), name: "e'ventname")
let data = try JSONEncoder().encode(event)
let entity = DataEntity(uniqueIdentifier: event.id.uuidString, timestamp: event.timestamp, data: data)

let res = queue.add(dataEntity: entity)  // This fails

```
#### Sample Code that illustrates the problem
https://www.sqlite.org/faq.html#q14

cdhoffmann:
Fixes issue #784 which is caused by including single quotes in entity data sent to SqliteDataQueue. As per: https://www.sqlite.org/faq.html#q14 which states that in order to escape single quotes in SQLite you simply add another single quote.


AEPServices/Tests/services/DataQueueTests.swift
```diff
@@ -535,4 +535,18 @@ class DataQueueTests: XCTestCase {
...
+        let queue = DataQueueService().getDataQueue(label: fileName)!
+        let event = EventEntity(id: UUID(), timestamp: Date(), name: "e'ventname")
+        let data = try JSONEncoder().encode(event)
+        let entity = DataEntity(uniqueIdentifier: event.id.uuidString, timestamp: event.timestamp, data: data)
+
+        // When
+        let result = queue.add(dataEntity: entity)
+
+        // Then
+        XCTAssertTrue(result)
```
kevinlind, line 550: Can you also validate that the data is still as expected after added to the database?  Basically validate that `e'ventname` is still the same?

cdhoffmann: Good call. Added

kevinlind (approved):

praveek (approved):

PR #799, opened 2022-04-28
[MOB-16309] Fixed an issue where initial Identity sync was not sent on app relaunch

addb:
This PR is based on PR: https://github.com/adobe/aepsdk-core-ios/pull/782 which needs to be merged before this.

## Types of changes

## Checklist:


AEPIdentity/Sources/IdentityState.swift
```diff
@@ -164,8 +165,11 @@ class IdentityState {
         identityProperties.mergeAndCleanCustomerIds(customerIds)
         customerIds.removeAll(where: { $0.identifier?.isEmpty ?? true }) // clean all identifiers by removing all that have a nil or empty identifier
 
+        // check is this is forceSync event. Currently the checks for event.forceSync is not very useful as there no events with that property set. It still here to support that workflow when there is a way to update the eventData.
```
emdobrin, line 168: rephrasing suggestion:
// Checks if this is forceSync event: either this is the first event processed by the extension ad boot time or if the eventData contains the forceSync flag for backwards compatibility.

AEPIdentity/Tests/IdentityStateTests.swift
```diff
@@ -639,6 +639,29 @@ class IdentityStateTests: XCTestCase {
         XCTAssertTrue(mockHitQueue.queuedHits.isEmpty) // hit should NOT be queued in the hit queue
     }
 
+    /// Tests that a hit is queued when forceSynce
+    func testSyncIdentifiers_whenForceSync_shouldQueueHit() {
```
emdobrin, line 643: Let's also add functional tests checking that shared states are updated as expected and the force sync network request is made on boot, on both install and subsequent launch scenarios.

emdobrin (changes requested):

AEPIdentity/Tests/IdentityStateTests.swift
```diff
@@ -639,6 +639,29 @@ class IdentityStateTests: XCTestCase {
...
+        var props = IdentityProperties()
+        props.ecid = ECID() // visitor ID is null initially and set for the first time in
+        // shouldSync(). Mimic a second call to shouldSync by setting the ecid
+        props.lastSync = Date() // set last sync to now
+        state = IdentityState(identityProperties: props, hitQueue: MockHitQueue(processor: MockHitProcessor()), pushIdManager: mockPushIdManager)
+        state.lastValidConfig = configSharedState
+
+        // test
+        let data = [IdentityConstants.EventDataKeys.IS_SYNC_EVENT: true] as [String: Any]
+        _ = state.syncIdentifiers(event: Event(name: "ID Sync Test Event", type: EventType.identity, source: EventSource.requestIdentity, data: data), forceSync: true)
```
kevinlind, line 657: You are setting both the event data `forcesync` flag and the function parameter `forceSync` to true which makes it hard to determine which flag is causing the forced sync. Instead, can you split this into two tests, one which sets the event data `forcesync` to true and the function parameter `forceSync` to false and another which does the opposite?

addb: @kevinlind I am setting IS_SYNC_EVENT eventdata flag and not FORCE_SYNC flag, but good catch I added one for FORCE_SYNC eventdata.

AEPIdentity/Tests/IdentityFunctionalTests.swift
```diff
@@ -51,6 +51,48 @@ class IdentityFunctionalTests: XCTestCase {
         XCTAssertNotNil(sharedState?[IdentityConstants.EventDataKeys.VISITOR_ID_ECID])
     }
 
+    /// Tests that a sync event dispatches a shared state update
+    func testSyncIdentifiersForceSyncEventInstallScenario() {
```
emdobrin, line 55: for these two new tests can we also check hit is queued / network request is sent?

addb: @emdobrin I was thinking of this but since we test directly the listeners first event will not be treated as forceSync since that is being called from isReadyForEvent. I can add a call to isReady and check for hit.

AEPIdentity/Sources/IdentityState.swift
```diff
@@ -16,15 +16,16 @@ import Foundation
 /// Manages the business logic of the Identity extension
 class IdentityState {
     private let LOG_TAG = "IdentityState"
-    private(set) var hitQueue: HitQueuing
     private var pushIdManager: PushIDManageable
     #if DEBUG
+        var hitQueue: HitQueuing
```
emdobrin, line 21: For functional testing can we use the network request mock instead? that will not require code changes and it will maintain the expected input/output format for functional tests.

AEPIdentity/Tests/IdentityFunctionalTests.swift
```diff
@@ -51,6 +73,61 @@ class IdentityFunctionalTests: XCTestCase {
...
+        // setup
+        setDefaultSuccessNetworkResponse()
+        let configSharedState = [IdentityConstants.Configuration.GLOBAL_CONFIG_PRIVACY: PrivacyStatus.optedIn.rawValue, IdentityConstants.Configuration.EXPERIENCE_CLOUD_ORGID: "test-org-id"]
+        let data = [IdentityConstants.EventDataKeys.IS_SYNC_EVENT: true]
+        let event = Event(name: "Sync Event", type: EventType.identity, source: EventSource.requestIdentity, data: data)
```
emdobrin, line 82: This input event seems invalid, can we have two tests instead:
1. input is getExperienceCloudID event https://github.com/adobe/aepsdk-core-ios/blob/main/AEPIdentity/Sources/Identity%2BPublicAPI.swift#L85-L88, output ECID returned and force sync network request sent
2. input is a valid syncIdentifier event - https://github.com/adobe/aepsdk-core-ios/blob/main/AEPIdentity/Sources/Identity%2BPublicAPI.swift#L128-L137, output first network request was sent for force sync with ecid, second request was sent with the new identity passed with the syncIdentifier event

AEPIdentity/Tests/IdentityFunctionalTests.swift
```diff
@@ -51,6 +73,61 @@ class IdentityFunctionalTests: XCTestCase {
...
+        mockRuntime.simulateSharedState(extensionName: "com.adobe.module.configuration", event: event, data: (configSharedState, .set))
+        waitForProcessing()
+
+        // test
+        XCTAssertTrue(identity.readyForEvent(event)) //// since first event will be processed by readyForEvent which calls forceSync
+        waitForProcessing()
```
emdobrin, line 88: This wait can be flaky as the processing time may defer on each testing machine, instead we should wait for the network request to execute to do the assertions.
I would recommend moving all these tests inside https://github.com/adobe/aepsdk-core-ios/blob/main/AEPIntegrationTests/IdentityIntegrationTests.swift and using the TestableNetworkService and the RequestResolver for this.

AEPIdentity/Tests/TestHelpers/MockNetworking.swift
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+import XCTest
+import AEPCore
+@testable import AEPIdentity
+import AEPCoreMocks
+import AEPServices
+import AEPServicesMocks
+
+class MockNetworking: Networking {
```
emdobrin, line 20: See my comment about moving the tests to Integration test, we can use https://github.com/adobe/aepsdk-core-ios/blob/main/AEPIntegrationTests/TestableNetworkService.swift

AEPIdentity/Tests/IdentityFunctionalTests.swift
```diff
@@ -51,6 +73,61 @@ class IdentityFunctionalTests: XCTestCase {
...
+        let sharedStateForInstall = mockRuntime.createdSharedStates.last!
+        XCTAssertNotNil(sharedStateForInstall?[IdentityConstants.EventDataKeys.VISITOR_ID_ECID])
+        XCTAssertTrue(mockNetworkService.connectAsyncCalled)
+        XCTAssertEqual(1, mockNetworkService.calledNetworkRequests.count)
+
+        let forceSyncdata = [IdentityConstants.EventDataKeys.IS_SYNC_EVENT: true, IdentityConstants.EventDataKeys.FORCE_SYNC: true]
+        let forceSyncEvent = Event(name: "Sync Event", type: EventType.identity, source: EventSource.requestIdentity, data: forceSyncdata)
+
+        // test
+        mockRuntime.simulateComingEvent(event: forceSyncEvent)
```
emdobrin, line 120: This scenario cannot be reproduced in a real setup since we don't have the FORCE_SYNC key inside event data. How about we transform this in integration test and simulate the SDK restart mid test:  extensions registration + config, check identity sends force sync, reset event hub and re-init extensions, check Identity sends a force sync
This might be helpful https://github.com/adobe/aepsdk-core-ios/blob/main/AEPIntegrationTests/LifecycleIntegrationTests.swift#L45

emdobrin (commented):
the new integration test you added looks good 🎉 , see my other comments on functional

AEPIdentity/Tests/IdentityFunctionalTests.swift
```diff
@@ -418,4 +430,8 @@ class IdentityFunctionalTests: XCTestCase {
         let ecid = dispatchedEvent?.data?[IdentityConstants.EventDataKeys.VISITOR_ID_ECID] as? String ?? ""
         XCTAssertFalse(ecid.isEmpty)
     }
+
+    func mockLastSyncTimeStamp(ts: Date = Date()) {
```
emdobrin, line 434: can we also remove mockLastSyncTimeStamp and waitForProcessing as they are not used?

emdobrin (approved):
Looks good!

AEPIdentity/Tests/IdentityFunctionalTests.swift
addb: oh yes. Thanks for catching it!

PR #800, opened 2022-04-29
[MOB-16285] Get latest non pending shared state

cdhoffmann:
In order to allow extensions to read cached config until the remote config is downloaded, a new API has been added to the `ExtensionRuntime` which allows us to fetch the latest `.set` shared state before an event. 

In order to support this new API, we have added a new parameter to the `ExtensionRuntime`'s `getSharedState` method, namely `resolution: SharedStateResolution` which is a new public Enum which currently holds the following values: `.lastSet` and `.any`. The API has a default parameter of resolution set to `.any` which will therefore not break compatibility. 

NOTE: If you are using a `TestableExtensionRuntime` in your Extension, you will need to update it to use the new api though. 

In order to read the cached config until the remote config is downloaded, simply set the `resolution` parameter to `.lastSet` in your extension. This will get the last set sharedState which will contain the cachedConfig.


AEPCore/Tests/FunctionalTests/EventHubContractTests.swift
```diff
@@ -376,7 +376,74 @@ class EventHubContractTest: XCTestCase {
...
+        MobileCore.dispatch(event: event1)
+        MobileCore.dispatch(event: event2)
+        MobileCore.registerExtensions([ContractExtensionOne.self]) {
+            startExpectation.fulfill()
+        }
+        wait(for: [startExpectation], timeout: 1)
+        let resolver = ContractExtensionOne.runtime?.createPendingSharedState(event: event1)
+
+        let sharedStateForEvent1 = ContractExtensionOne.runtime?.getSharedState(extensionName: "com.adobe.ContractExtensionOne", event: event1, barrier: true, resolution: .lastSet)
+        XCTAssertEqual(SharedStateStatus.set, sharedStateForEvent1?.status)
```
kevinlind, line 433: Why is the returned status `.set` here?  If event1 has a pending shared state and that is the only shared state, shouldn't `.lastSet` resolve to `SharedStateStatus.none`?

kevinlind (commented):
Can you add test cases to cover the lines marked by Codecov as not covered?

AEPCore/Tests/FunctionalTests/EventHubContractTests.swift
cdhoffmann: Good catch. I should be returning .none from the SharedState.resolve

kevinlind (approved):
Looks good.

kevinlind (approved):

pfransenadb (approved):
👍

Issue #797, opened 2022-04-26
Support bundled rules


praveek:
Core downloads Launch rules from the remote https endpoint for "rules.url" key in configuration. 

On first launch, Configuration extension downloads the rules bundle and caches it locally. 
On subsequent launches, the cached rules are loaded first before attempting to check if the remote rules have changed. 

This workflow causes an issue, especially for first launch, where rules are not precessed for the events dispatched before download completes. We handle this scenario by caching events dispatched before downloading the rules and processing it again after download. However, this does not handle rules which change current event data (“add” and “modify” consequences)

To workaround this issue, we can allow developers to download and bundle rules with the app. The SDK should look for a file named `ADBMobileConfig-rules.zip` as the default configuration is expected to be `ADBMobileConfig.json`. 

- On launch, Configuration extension will try to load cached rules. 
- If no cached rules are present, it checks and loads bundled rules (ADBMobileConfig-rules.zip) if present. 
- It will then attempt to download remote rules from Configuration.

PR #801, opened 2022-05-06
[PLATIR-19945] Bundled rules support

cdhoffmann:
Support for bundled rules as specified in #797. Configuration will first check for cached rules, if cached rules are not present (first launch scenario) we will use bundled rules if they are found.


pfransenadb (approved):
Looks good to me, simpler than I thought it was going to be. Nicely done. 👍

Issue #707, opened 2021-09-01
Add tvOS Support 


praveek:
- Add tvOS support to AEPServices
- Add tvOS support for Core extensions
- Add tvOS target to sample apps

mmcnabb2187:
We would like to see this prioritized.  We have a number of projects that need to update and are finding it problematic to use the static ACP* libraries.  Thank you.

objectiveCarlo:
Hello are there any updates on this issue? Thank you.

praveek:
@objectiveCarlo We are planning to add tvOS support by end of July.

alexzchut:
@praveek I am using AEP with iOS and need this also for tvOS to be used with several client's apps so I prepared PRs with changes needed to allow tvOS support, 
in core one - preventing access to iOS only related stuff, on others just adding tvOS in podspec. 
I see that events are being sent correctly so please take a look if this can help to release tvOS version sooner.

Thanks!
 
https://github.com/adobe/aepsdk-core-ios/pull/798
https://github.com/adobe/aepsdk-rulesengine-ios/pull/52
https://github.com/adobe/aepsdk-analytics-ios/pull/69
https://github.com/adobe/aepsdk-media-ios/pull/53

praveek:
Thanks for the contribution @alexzchut. We have planned to release tvOS support for Core, Analytics and Media around July. As part of the official release, we are also working on updating few services, sample apps, allow using it from cocoapods, spm and also certifying the builds on device. 

cc @praschetan

wryphonedev:
@praveek Any chance we'll see this merged and released soon? As we try to adopt Analytics on iOS and tvOS, we either need this published solution or we have to diverge our project and have tv integrate a much older sdk.

PR #811, opened 2022-06-03
Finish tvOS Support

cdhoffmann:
Adds full support for tvOS to Core. #707


praveek:
LGTM. Can you also update MakeFile and Circle CI steps to build/test for tvOS?

Package.resolved
```diff
@@ -0,0 +1,16 @@
+{
```
addb, line 1: Do we need to push this file to git?

Podfile
```diff
@@ -73,3 +73,21 @@ end
...
 end
+
+target 'TestApp_Swift (tvOS)' do
+  tests_main
+end
+
+target 'TestApp_Objc (tvOS)' do
+  tests_main
+end
+
```
addb, line 84: Should we add tvOS to the platform to the top of this file? https://github.com/adobe/aepsdk-core-ios/pull/811/files#diff-8f7d6adf31268a2d897ee34bd170592648d6e520aa237104395e4a4438af50cbR2

cdhoffmann: Sorry that link isn't taking me to a specific file. Which file?

addb: we have platform field at the top of the podfile L2

cdhoffmann: Oh, you can only have one platform. I am not really sure how that affects things here to be honest.

addb: I think we need to remove this line and specify the platform and version for each target.

addb: @cdhoffmann I checked with our ACP extensions which have TvOS support, it looks like L2 is fine haveing one platform but we need to add TvOS platform and version for TvOS targets.

praveek (approved):

addb (approved):

Podfile
praveek: I am not sure, but i think Cocoapod links the correct dependency based on the target platform. I also think specifying the version checks whether all the dependencies satisfy the condition.

Issue #712, opened 2021-09-03
Investigate how Core extensions handle offline scenarios 


praveek:
- Investigate how Core extensions work when device is offline. 
- Configuration currently keeps retrying the endpoint every few seconds when the device offline. Investigate about increasing the timeout after few failures.  
- Provide a recommendation on how other extensions should handle offline cases.

PR #819, opened 2022-06-24
Configuration retry improvements

cdhoffmann:
#712 Improves configuration retry logic by sending a configuration request event on retry instead of recursively calling the processor. Also scales the retry by increasing the retry deadline after every retry.

### DIFF
--- a/AEPCore/Sources/configuration/Configuration.swift
+++ b/AEPCore/Sources/configuration/Configuration.swift
@@ -148,6 +148,7 @@ class Configuration: NSObject, Extension {
                 guard let self = self else { return }
                 Log.trace(label: self.name, "Downloading config failed, trying again")
                 self.retryQueue.asyncAfter(deadline: .now() + (self.retryConfigurationCounter * 2)) {
+                    print("retrying: \(self.retryConfigurationCounter)")
                     let event = Event(name: CoreConstants.EventNames.CONFIGURE_WITH_APP_ID, type: EventType.configuration, source: EventSource.requestContent,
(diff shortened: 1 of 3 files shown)


AEPCore/Sources/configuration/Configuration.swift
```diff
@@ -146,8 +147,12 @@ class Configuration: NSObject, Extension {
                 // If downloading config failed try again later
                 guard let self = self else { return }
                 Log.trace(label: self.name, "Downloading config failed, trying again")
-                self.retryQueue.asyncAfter(deadline: .now() + 5) {
-                    self.processConfigureWith(appId: appId, event: event, sharedStateResolver: sharedStateResolver)
+                self.retryQueue.asyncAfter(deadline: .now() + (self.retryConfigurationCounter * 2)) {
```
praveek, line 150: Can you resolve the pending shared state with current config before retrying the network request and add a test around it?

praveek, line 150: Can we change it to `self.retryConfigurationCounter * 5` to increase the time between initial retries.

AEPCore/Sources/configuration/Configuration.swift
```diff
@@ -146,8 +147,12 @@ class Configuration: NSObject, Extension {
                 // If downloading config failed try again later
                 guard let self = self else { return }
                 Log.trace(label: self.name, "Downloading config failed, trying again")
```
praveek, line 149: nit: trying again after xx secs

praveek (approved):

PR #824, opened 2022-07-13
[MOB-16731] Boot up Identity on registration

addb:
[MOB-16731] Boot Identity state in extension onRegistered method to speed up getters.

## Types of changes

## Checklist:


AEPIdentity/Sources/Identity.swift
```diff
@@ -58,7 +60,7 @@ import Foundation
         guard let state = state else { return false }
 
         // fast boot identity without waiting for configuration
-        state.boot(event: event, createSharedState: createSharedState(data:event:))
+        state.boot(createSharedState: createSharedState(data:event:))
```
praveek, line 63: I don't think this is needed as you already do it in the `onRegistered` method.

addb: Yes, I was thinking about the case where state is null onRegistered but yes then it would be null everywhere so doesn't matter. I will remove it.

emdobrin (commented):
@addb changes look good. 
One comment is can we add some tests for the launch/install scenarios and measure the time of the operation so we don't miss this with incremental updates in the future? You may need to use a threshold for the execution time, but it should not take more than 0.5 seconds for example.

AEPIdentity/Sources/IdentityState.swift
```diff
@@ -44,15 +44,14 @@ class IdentityState {
...
         if hasBooted { return }
 
         // load data from local storage
         identityProperties.loadFromPersistence()
 
         if identityProperties.ecid != nil {
-            createSharedState(identityProperties.toEventData(), event)
+            createSharedState(identityProperties.toEventData(), nil)
             didCreateInitialSharedState = true
         }
```
praveek, line 57: Can you remove the `hasBooted` variable as it is not used anywhere?

praveek (approved):

AEPIdentity/Sources/IdentityState.swift
emdobrin: we use it to verify this function ran only once when the extension was first initialized, to prevent multiple calls to it.

emdobrin (approved):

Issue #814, opened 2022-06-13
Investigate using Xcode 13 in CI


cdhoffmann:
Make sure that using Xcode 13 works for CircleCI and release scripts

PR #826, opened 2022-07-15
Update CircleCI to Xcode 13

cdhoffmann:
#814 Updates CircleCI Xcode version to Xcode 13. Minimum version to fully support XCFrameworks for all customers.


praveek (approved):

praveek:
Do you know why the code coverage numbers dropped with Xcode version? Should we enable any other setting?

cdhoffmann:
@praveek I hadn't noticed that the coverage went down. I don't think there is any other setting. Let me look into why that happened.

emdobrin:
We noticed a similar behavior for the other repos after upgrading to xcode 13, seems related to https://developer.apple.com/forums/thread/690991

PR #828, opened 2022-08-01
[PLATIR-22271] Fix rules processing delay

cdhoffmann:
Our `replaceRules` function was async which was causing a dsync between rules processing, and setting the rules. Events which come in early were mostly affected as rules processing occurs on the event before the replaceRules async call finishes. 

Also added better logging in order to more easily debug issues with rules in the future.


AEPCore/Sources/eventhub/EventHub.swift
```diff
@@ -48,7 +48,7 @@ final class EventHub {
         // Setup eventQueue handler for the main OperationOrderer
         eventQueue.setHandler { [weak self] (event) -> Bool in
             guard let processedEvent = self?.preprocessors.shallowCopy.reduce(event, {$1($0)}) else { return true }
-
+            Log.debug(label: self?.LOG_TAG ?? "EventHub", "Processed event: \(processedEvent)")
```
praveek, line 51: Can you change this to trace and also add additional check as it lies in hot path?
if Log.logFilter == .trace {
  Log.trace(label: self?.LOG_TAG ?? "EventHub",
                      "Processed Event #\(String(describing: self?.eventNumberMap[event.id] ?? 0)) - \(processedEvent)")
}

praveek: Can you also add a similar check in dispatch(event: Event) function

praveek (approved):

AEPCore/Sources/eventhub/EventHub.swift
cdhoffmann: Why do we need the additional check? The Log.trace function does the check in the function.

AEPCore/Sources/rules/LaunchRulesEngine.swift
```diff
@@ -175,7 +175,7 @@ public class LaunchRulesEngine {
             Log.error(label: LOG_TAG, "(\(self.name)) : Unable to process an AttachDataConsequence Event, 'eventData' is missing from original event")
             return nil
         }
-        Log.trace(label: LOG_TAG, "(\(self.name)) : Attaching event data with \(PrettyDictionary.prettify(from))\n")
+        Log.trace(label: LOG_TAG, "(\(self.name)) : Attaching event data: \(PrettyDictionary.prettify(from)) to \(PrettyDictionary.prettify(eventData))\n")
```
praveek, line 178: nit: Use to variable instead of eventData

AEPCore/Sources/rules/LaunchRulesEngine.swift
```diff
@@ -194,7 +194,7 @@ public class LaunchRulesEngine {
             Log.error(label: LOG_TAG, "(\(self.name)) : Unable to process a ModifyDataConsequence Event, 'eventData' is missing from original event")
             return nil
         }
-        Log.trace(label: LOG_TAG, "(\(self.name)) : Modifying event data with \(PrettyDictionary.prettify(from))\n")
+        Log.trace(label: LOG_TAG, "(\(self.name)) : Modifying event data: \(PrettyDictionary.prettify(eventData)) with data: \(PrettyDictionary.prettify(from))\n")
```
praveek, line 197: nit: Use to variable instead of eventData

praveek (approved):

PR #853, opened 2022-11-08
[MOB-17595], [MOB-17596] - don't block mt and skip delegate check on public show calls

sbenedicadb:


AEPServices/Sources/ui/fullscreen/FullscreenMessage.swift
```diff
@@ -151,17 +148,27 @@
...
+                return
+            }
+
+            // dispatch UI activity back to main thread
+            DispatchQueue.main.async {
                 // notify global listeners
-                self.listener?.onShow(message: self)
-                self.messagingDelegate?.onShow(message: self)
+                message.listener?.onShow(message: message)
+                message.messagingDelegate?.onShow(message: message)
```
pfransenadb, line 169: This is still customer controlled code on main thread (potentially) ?

It's confusing, but should we consider dispatching this off to global as well?  would that cause a problem if listener.onshow and messagingDelegate.onShow happened non-linearly (or -- more specifically -- farther apart in a non blocking fashion before message.displayWithAnimation ?)

pfransenadb (commented):
Question about main thread of onShow for message delegate.

cdhoffmann (approved):

PR #855, opened 2022-11-09
[MOB-17616] Make our IdentityProperties merging safe from crashes.

cdhoffmann:
Currently our IdentityProperties uses an unsafe Apple API `Dictionary(uniqueKeysWithValues:)` which throws an exception if duplicate keys are fed into it. Although we don't expect two CustomerIds to have the same type, a crash will happen if they do. 

Fix: Use uniquingKeysWith API instead and pick the second value if there are duplicates.


TestProject/.gitignore
```diff
@@ -0,0 +1,9 @@
+.DS_Store
```
praveek, line 1: Is the test project checked in by mistake?

cdhoffmann: No idea where that came from. Removed.

praveek (approved):

praveek:
Not related to your change but can you add these three zip files to `GitIgnore` so that they don't get commited each time.

sbenedicadb (approved):

PR #856, opened 2022-11-11
bump version for 3.7.3 release

cdhoffmann:
Bump version for 3.7.3 release which fixes a bug in Identity


sbenedicadb (approved):

sbenedicadb:
let's write a script to update the version in all these files so we can be more confident that we didn't miss any!

PR #866, opened 2022-12-12
Version Updater GH Action

sbenedicadb:
an easy way to update all the versions references using a github action


Script/update-versions.sh
```diff
@@ -0,0 +1,121 @@
+#!/bin/bash
```
praveek, line 1: Lots of shell script 😨

praveek (approved):
Thanks for automating this step ❤️

Issue #874, opened 2023-01-12
ServiceProvider.shared.loggingService cannot be overridden


m0rk:
### Prerequisites

- [X] This is not a Security Disclosure, otherwise please follow the guidelines in [Security Policy](https://github.com/adobe/aepsdk-edge-ios/security/policy).
- [X] I have searched in this repository's issues to see if it has already been reported.
- [X] I have updated to the latest released version of the SDK and the issue still persists.

### Bug summary

According to the [documentation](https://github.com/adobe/aepsdk-core-ios/tree/main/Documentation/Services#overriding-a-service) it should be possible to override the logging service, but the [implementation](https://github.com/adobe/aepsdk-core-ios/blob/main/AEPServices/Sources/ServiceProvider.swift#L101-L105) is a get-only property.

### Environment

* OS: iOS 16.1
* SDK(s): AEPCore 3.7.3
* IDE: Xcode 14.1

cdhoffmann:
@m0rk It is actually intentional for the logging service to not be overridable. We do not want customers overriding it because we need our logs when helping them track down issues. The documentation is misleading though. We will make sure and update the documentation to not use the logging service as an example for overriding. Thank you.

PR #875, opened 2023-01-12
Make it possible to override the loggingService

m0rk:
## Description

According to the [documentation](https://github.com/adobe/aepsdk-core-ios/tree/main/Documentation/Services#overriding-a-service) this should be possible, but the implementation was a get-only property.

## Related Issue

https://github.com/adobe/aepsdk-core-ios/issues/874

## Motivation and Context

First of all it is mentioned as an example in the documentation of the ServiceProvider but the main reason is that we have a use case for writing the logs somewhere else for automated testing purposes.

## How Has This Been Tested?

I wrote a unit test.

## Types of changes

I am not entirely sure how to categorise this. If we take the documentation as the source of truth it would be a bug, else it is a non-breaking new feature. 

## Checklist:

- [x] I have signed the [Adobe Open Source CLA](http://opensource.adobe.com/cla.html).
- [x] My code follows the code style of this project.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.


cdhoffmann:
@m0rk Can you sign Adobe's CLA? I will merge it once you do. Thanks for the contribution!

m0rk:
@cdhoffmann I am a bit confused now, according to the last comment on the related issue you did not want customers overriding it, was there a change of heart in this regard? 😄 If this is indeed the case I will get the CLA resolved as fast as possible.

PS: Sorry for the late reply, I was sick.

cdhoffmann:
@m0rk I apologize for the confusion. Originally that was our stance, but we have recently made this same change in our Android SDK so it will line up now. Thanks again for your contribution!

m0rk:
Hello @cdhoffmann, I finally got the CLA signed.

cdhoffmann (approved):

PR #885, opened 2023-03-07
Add EventType and EventSource for Edge Media

kevinlind:
## Description

Adding EventType and EventSource for Edge Media.

See also Android Core PR with similar change: https://github.com/adobe/aepsdk-core-android/pull/403

## Types of changes

## Checklist:


addb (approved):
looks good to me!

AEPCore/Sources/eventhub/EventSource.swift
```diff
@@ -36,4 +36,8 @@ public class EventSource: NSObject {
     public static let applicationClose = "com.adobe.eventSource.applicationClose"
     public static let personalizationDecisions = "personalization:decisions"
     public static let locationHintResult = "locationHint:result"
+    public static let errorResponseContent = "com.adobe.eventSource.errorResponseContent";
+    public static let requestTracker = "com.adobe.eventSource.requestTracker";
+    public static let trackMedia = "com.adobe.eventSource.trackMedia";
+    public static let sessionCreated = "com.adobe.eventSource.sessionCreated";
```
praveek, line 42: We previously dispatched a similar event in media to match client API events with hitdebugger response. Does this still apply to edge media?

AEPCore/Sources/eventhub/EventSource.swift
```diff
@@ -36,4 +36,8 @@ public class EventSource: NSObject {
     public static let applicationClose = "com.adobe.eventSource.applicationClose"
     public static let personalizationDecisions = "personalization:decisions"
     public static let locationHintResult = "locationHint:result"
+    public static let errorResponseContent = "com.adobe.eventSource.errorResponseContent";
+    public static let requestTracker = "com.adobe.eventSource.requestTracker";
```
praveek, line 40: nit: Since media doesn't send a corresponding response event, should we use createTracker instead?

praveek:
Are there plans to update media direct extension to use the new event type/source?

kevinlind:
> Are there plans to update media direct extension to use the new eventsource?

We've previously argued that event are considered part of the public API. The media direct extensions use a naming convention which included the extension name in the event source string ("com.adobe.eventsource.media.trackmedia"). I would like to fix that with this PR, but then we would not be able to update the media direct extensions without the risk of breaking someone's implementation if they are listening for these events.

AEPCore/Sources/eventhub/EventSource.swift
addb: We have kept this for now. We don't have the integration with hitdebugger for now, but I am thinking we might need this for low level events validation in griffon.

emdobrin:
I agree with @kevinlind that we should not modify the event types/sources for Media direct as it may be breaking some implementations.

AEPCore/Sources/eventhub/EventSource.swift
emdobrin: @addb there will likely be no client-side integration with hitdebugger in the case of Edge Media. For debugging purposes can we relay exclusively on the `media-analytics:new-session` response handle and not dispatch this additional internal event?

kevinlind:
Closing PR. We'll keep the type/source internal to Edge Media for now and reevaluate the need for this PR closer to the Edge Media GM.

PR #917, opened 2023-04-27
Make circleCI jobs async

cdhoffmann:
Makes circle ci jobs async to cut down on time

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,224 +3,363 @@
 
 version: 2.1
 
+orbs:
+  codecov: codecov/codecov@3.2.4
+  macos: circleci/macos@2
+  
 # Workflows orchestrate a set of jobs to be run;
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -3,140 +3,363 @@
...
 # Workflows orchestrate a set of jobs to be run;
 workflows:
   version: 2
   build-test:
     jobs:
-      - build-and-test
+      - validate-code
+      - test-ios-core:
+          requires:
+            - validate-code
```
praveek, line 18: Can we remove this from each job and just run it once? I see that `validate-code` runs lint for all the extensions.

cdhoffmann: @praveek Yeah that only runs once. The `requires` is there to let the workflow know that it needs to run before all the other jobs run.

PR #926, opened 2023-05-11
Fix event dispatch when creating shared state

praveek:
Dispatch event in order when creating shared state.
Record event history after processing rules for event.


cdhoffmann (changes requested):
Can we add tests for these two cases?

cdhoffmann (approved):

PR #937, opened 2023-05-30
Min version bump to iOS 11.0

praveek:
Fix most Xcode 14 warnings.

Updates the minimum deployment version for iOS and tvOS to 11.0 as Xcode 14 drops v10 support.
Turns of the dependency analysis flag for our https://github.com/realm/SwiftLint/issues/4015.
There is still a warning about SwiftLint's minimum supported version being iOS 9, this is because SwiftLint has not updated their podspec.

### DIFF
--- a/AEPCore.xcodeproj/project.pbxproj
+++ b/AEPCore.xcodeproj/project.pbxproj
@@ -3,7 +3,7 @@
 	archiveVersion = 1;
 	classes = {
 	};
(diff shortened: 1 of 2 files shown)


cdhoffmann (commented):
We're missing the script change for Swift Lint

cdhoffmann (approved):

PR #968, opened 2023-09-21
[MOB-19513] Migration logic

cdhoffmann:
The migration logic for migrating data from UserDefaults to new datastore

### DIFF
--- a/AEPCore/Sources/migration/UserDefaultsMigrator.swift
+++ b/AEPCore/Sources/migration/UserDefaultsMigrator.swift
@@ -37,13 +37,18 @@ struct UserDefaultsMigrator {
                 }
                 Log.debug(label: LOG_TAG, "UserDefaults Migration complete for \(collectionName)")
             }
+            dataStore.set(collectionName: constants.MIGRATION_STORE_NAME, key: constants.MIGRATION_COMPLETE, value: true)
             Log.debug(label: LOG_TAG, "UserDefaults migration complete")
         }
     }
     
     private func needToMigrate() -> Bool {
-        let installDateKey = keyWithPrefix(datastoreName: constants.Lifecycle.DATASTORE_NAME, key: constants.Lifecycle.DataStoreKeys.INSTALL_DATE.rawValue)
-        return defaults.object(forKey: installDateKey) != nil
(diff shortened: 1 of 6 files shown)


AEPCore/Sources/migration/UserDefaultsMigrator.swift
```diff
@@ -0,0 +1,62 @@
...
+                    }
+                }
+                Log.debug(label: LOG_TAG, "UserDefaults Migration complete for \(collectionName)")
+            }
+            Log.debug(label: LOG_TAG, "UserDefaults migration complete")
+        }
+    }
+    
+    private func needToMigrate() -> Bool {
+        let installDateKey = keyWithPrefix(datastoreName: constants.Lifecycle.DATASTORE_NAME, key: constants.Lifecycle.DataStoreKeys.INSTALL_DATE.rawValue)
```
praveek, line 45: We should not do the migration based on this key as it will be present only for customers using Lifecycle extension. 
I can't think of a key to use as a basis for the migration as even our config keys won't be available when using bundled configuration. Can you run the migration logic with out the `needToMigrate` check as it should become a no-op after the first run?

cdhoffmann: I was following the pattern we use for the other migrators. Do you think we should add a key to our DataStore post migration that we can check later to avoid running through the loops on every launch?

AEPCore/Sources/migration/UserDefaultsMigrator.swift
```diff
@@ -0,0 +1,62 @@
...
+        return defaults.object(forKey: installDateKey) != nil
+    }
+    
+    private func keyWithPrefix(datastoreName: String, key: String) -> String {
+        return "Adobe.\(datastoreName).\(key)"
+    }
+    
+    private func getAndDelete(key: String) -> Any? {
+        guard let value = defaults.object(forKey: key) else {
+            Log.debug(label: LOG_TAG, "Failed to get value to migrate from UserDefaults")
```
pfransenadb, line 55: nit: I'd prefer to use language other than "failed" here -- as we fully expect a bunch of the keys to not exist (Depending on implementation).  Can we change to a verbose log and change verbiage to "No value for \(key) found."

pfransenadb (changes requested):
Minor nit/changes requested around logging in migrator.

pfransenadb (approved):
Looks good 👍

Issue #954, opened 2023-07-25
Investigate random failures in integration tests


sbenedicadb:
### Task description

Details can be found in this workflow: https://app.circleci.com/pipelines/github/adobe/aepsdk-core-ios/2188/workflows/4f6fd9bd-1b28-4051-9d5f-85e24ef4942e

cdhoffmann:
Another failing test: ConfigurationIntegrationTests.testClearUpdatedConfig()

https://app.circleci.com/pipelines/github/adobe/aepsdk-core-ios/2236/workflows/d29f3d50-76c4-4107-8b5b-1e083d4d79dd/jobs/4075

PR #969, opened 2023-09-22
Modify cache to not read file modification date

praveek:
- Modify CacheService to stop using fileModificationDate attribute to store cacheExpiry. 
- Clean up logs in FileSystemNamedCollection.

This change will cause a one time cache miss for currently cached content. 

Fixed `IdentityIntegrationTests .testGetExperienceCloudIdWithinPermissibleTimeOnInstall` - https://github.com/adobe/aepsdk-core-ios/issues/954

### DIFF
--- a/AEPServices/Sources/storage/FileSystemNamedCollection.swift
+++ b/AEPServices/Sources/storage/FileSystemNamedCollection.swift
@@ -49,7 +49,7 @@ class FileSystemNamedCollection: NamedCollectionProcessing {
                 do {
                     try updatedStorageData.write(to: fileUrl, options: .atomic)
                 } catch {
-                    Log.warning(label: self.LOG_TAG, "Error \(error) when writing \(key) to collection \(collectionName)")
+                    Log.warning(label: self.LOG_TAG, "Error '\(error)' when writing '\(key)' to collection '\(collectionName)'")
                 }
             }
         }
@@ -75,7 +75,7 @@ class FileSystemNamedCollection: NamedCollectionProcessing {
                     do {
                         try updatedStorageData.write(to: fileUrl, options: .atomic)
                     } catch {
-                        Log.warning(label: self.LOG_TAG, "Error \(error) when attempting to remove key \(key) from collection \(collectionName)")
+                        Log.warning(label: self.LOG_TAG, "Error '\(error)' when attempting to remove key '\(key)' from collection '\(collectionName)'")
                     }
                 }
             }
@@ -97,7 +97,7 @@ class FileSystemNamedCollection: NamedCollectionProcessing {
         }
 
         guard let jsonResult = try? JSONSerialization.jsonObject(with: storageData) as? [String: Any] else {
-            Log.warning(label: LOG_TAG, "Failed to read dictionary from collection \(collectionName)")
+            Log.warning(label: LOG_TAG, "Failed to read dictionary from collection '\(collectionName)'")
             return nil
         }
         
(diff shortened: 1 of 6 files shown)


AEPServices/Sources/storage/FileSystemNamedCollection.swift
```diff
@@ -49,7 +49,7 @@ class FileSystemNamedCollection: NamedCollectionProcessing {
                 do {
                     try updatedStorageData.write(to: fileUrl, options: .atomic)
                 } catch {
-                    Log.warning(label: self.LOG_TAG, "Error when writing to file: \(error)")
+                    Log.warning(label: self.LOG_TAG, "Error \(error) when writing \(key) to collection \(collectionName)")
```
sbenedicadb, line 52: i would vote to put our dynamic content in single quotes so it's clear to the reader exactly where the error, key, and collection name begin and end.  should we also announce `key` (e.g. - `...when writing key '\(key)'`) like we are for the other variables and in the subsequent log?

AEPServices/Sources/storage/FileSystemNamedCollection.swift
```diff
@@ -92,13 +92,16 @@ class FileSystemNamedCollection: NamedCollectionProcessing {
...
+        }
+
+        guard let jsonResult = try? JSONSerialization.jsonObject(with: storageData) as? [String: Any] else {
+            Log.warning(label: LOG_TAG, "Failed to read dictionary from collection \(collectionName)")
+            return nil
         }
-        Log.warning(label: LOG_TAG, "Failed to get Dictionary from data store")
-        return nil
+        
+        return jsonResult
```
sbenedicadb, line 104: 👍

AEPServices/Sources/storage/FileSystemNamedCollection.swift
```diff
@@ -92,13 +92,16 @@ class FileSystemNamedCollection: NamedCollectionProcessing {
             return nil
         }
 
-        if let storageData = try? Data(contentsOf: fileUrl) {
-            if let jsonResult = try? JSONSerialization.jsonObject(with: storageData) as? [String: Any] {
-                return jsonResult
-            }
+        guard let storageData = try? Data(contentsOf: fileUrl) else {
+            return nil
```
sbenedicadb, line 96: should we provide a message here indicating we were unable to retrieve the file contents?

AEPServices/Tests/services/DiskCacheServiceTests.swift
```diff
@@ -184,3 +215,16 @@ class DiskCacheServiceTests: XCTestCase {
...
 }
+
+private extension Date {
+    func equal(other: Date?) -> Bool {
+        guard let other = other else {
+            return false;
+        }
+
+        // As date stores the timestamp in milliseconds, compare double value with accuracy greater than milliseconds.
+        let accuracy = 0.00001;
```
sbenedicadb, line 226: wouldn't this just be 0.001?  my two cents is we could make it even a little more forgiving than this so we don't have random failures with this equality check

sbenedicadb (approved):
couple small things - take them or leave them

AEPServices/Sources/storage/FileSystemNamedCollection.swift
praveek: This is causing a lot of false warnings during a new install.

cdhoffmann (approved):

PR #971, opened 2023-10-16
Add attribute for fileprotection

cdhoffmann:
Adds an attribute for our FileSystem storage directory and sets the fileprotection to none. This is to avoid system restrictions during app backgrounding and other potential edge cases.

Also adds capabilities to our test app for testing backgrounding tasks.

### DIFF
--- a/AEPServices/Sources/storage/FileSystemNamedCollection.swift
+++ b/AEPServices/Sources/storage/FileSystemNamedCollection.swift
@@ -126,6 +126,7 @@ class FileSystemNamedCollection: NamedCollectionProcessing {
             let adobeBaseUrl = baseUrl.appendingPathComponent(adobeDirectory, isDirectory: true)
             do {
                 try fileManager.createDirectory(at: adobeBaseUrl, withIntermediateDirectories: true)
+                try fileManager.setAttributes([FileAttributeKey.protectionKey: FileProtectionType.none], ofItemAtPath: adobeBaseUrl.path)
             } catch {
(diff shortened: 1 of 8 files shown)


AEPServices/Sources/storage/FileSystemNamedCollection.swift
```diff
@@ -126,6 +126,7 @@ class FileSystemNamedCollection: NamedCollectionProcessing {
             let adobeBaseUrl = baseUrl.appendingPathComponent(adobeDirectory, isDirectory: true)
             do {
                 try fileManager.createDirectory(at: adobeBaseUrl, withIntermediateDirectories: true)
+                try fileManager.setAttributes([FileAttributeKey.protectionKey: FileProtectionType.none], ofItemAtPath: adobeBaseUrl.path)
             } catch {
                 Log.error(label: adobeDirectory, "Failed to create storage directory with error: \(error)")
```
praveek, line 131: Should we consider adjusting this error message? Is it possible to check the reason and avoid printing this message when the error is a result of the directory already existing?

cdhoffmann: I believe it doesn't error when the directory already exists

cdhoffmann: https://github.com/adobe/aepsdk-core-ios/pull/966#discussion_r1318920171

praveek (approved):

PR #974, opened 2023-10-25
[MOB-19790] Fix encoded data in migration

cdhoffmann:
Fixes crash which happened when trying to set Data in FileSystemNamedCollection. 

The crash occurred because the JSONSerialization.data(:_) method crashes internally when a top level fragment is not an accepted type. See [here](https://developer.apple.com/documentation/foundation/jsonserialization). When we attempted to set encoded data, the crash occurs. 

To avoid the crash, we use the `JSONSerialization.isValidJSONObject` to check if the value can be used. 
To successfully migrate the encoded data, we now encode the data as a string in the migrator before migrating it.

### DIFF
--- a/AEPCore/Sources/migration/UserDefaultsMigrator.swift
+++ b/AEPCore/Sources/migration/UserDefaultsMigrator.swift
@@ -31,7 +31,10 @@ struct UserDefaultsMigrator {
             Log.debug(label: LOG_TAG, "Beginning UserDefaults migration")
             for (collectionName, keys) in constants.migrationDict {
                 for key in keys {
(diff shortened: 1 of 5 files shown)


kevinlind (approved):
Looks good!
Were you able to test with Core's test app? Did you see any cases where the migrated data was not serializable?

praveek (approved):

cdhoffmann:
@kevinlind Just tested on the core test app and the issue was that we were setting the app group on the dev branch but not on main. I went ahead and removed that from our dev branch to avoid issues.

PR #982, opened 2023-12-19
Reverts to use UserDefaultsNamedCollection on tvOS

cdhoffmann:
tvOS does not support FileSystem storage for persistence. Works on simulator but not on a real device.


praveek:
Can you also wrap the [following logic](https://github.com/adobe/aepsdk-core-ios/blob/main/AEPCore/Sources/core/MobileCore.swift#L49) so that it executes only for iOS?

praveek (approved):

Issue #986, opened 2024-01-05
Extend rulesengine to support token replacement in arrays


praveek:
### Prerequisites

- [X] This is not a Security Disclosure, otherwise please follow the guidelines in [Security Policy](https://github.com/adobe/aepsdk-edge-ios/security/policy).
- [X] I have searched in this repository's issues to see if it has already been reported.

### Feature request summary

The current implementation of the Rules Engine only supports token replacement in Strings or Maps. Extend it to support arrays.

PR #988, opened 2024-01-16
[PLATIR-35300] Add array replace token support

cdhoffmann:
#986 Adds array string token replacement support for the LaunchRulesEngine

### DIFF
--- a/AEPCore/Sources/migration/UserDefaultMigrationConstants.swift
+++ b/AEPCore/Sources/migration/UserDefaultMigrationConstants.swift
@@ -9,7 +9,7 @@
  the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
  OF ANY KIND, either express or implied. See the License for the specific language
  governing permissions and limitations under the License.
-*/
+ */
 
 import Foundation
 
@@ -177,7 +177,7 @@ enum UserDefaultMigratorConstants {
         static let DATASTORE_NAME = "com.adobe.edge.identity"
         
         enum DataStoreKeys: String, CaseIterable {
-           case IDENTITY_PROPERTIES = "identity.properties"
+            case IDENTITY_PROPERTIES = "identity.properties"
         }
     }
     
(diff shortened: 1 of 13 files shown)


AEPCore/Sources/migration/UserDefaultMigrationConstants.swift
```diff
@@ -177,15 +177,15 @@ enum UserDefaultMigratorConstants {
         static let DATASTORE_NAME = "com.adobe.edge.identity"
         
         enum DataStoreKeys: String, CaseIterable {
-           case IDENTITY_PROPERTIES = "identity.properties"
+            case IDENTITY_PROPERTIES = "identity.properties"
```
praveek, line 180: Did we commit unformatted files earlier, or is there any change in Swift format version that we're using?

AEPCore/Sources/rules/LaunchRulesEngine.swift
```diff
@@ -293,6 +293,16 @@ public class LaunchRulesEngine {
             case is [String: Any]:
                 let valueDict = mutableDict[key] as! [String: Any]
                 mutableDict[key] = replaceToken(in: valueDict, data: data)
+            case is [Any]:
+                let valueArr = mutableDict[key] as! [Any]
```
praveek, line 297: Can we extend support for nested token replacement in arrays to maintain consistency with maps?

AEPCore/Sources/migration/UserDefaultMigrationConstants.swift
cdhoffmann: I'm not too sure how this happened. I just ran swiftformat locally and committed.

AEPCore/Sources/rules/LaunchRulesEngine.swift
cdhoffmann: Good call. I have added support and made it more generic to handle nested dictionaries inside nested arrays and vice versa.

AEPCore/Sources/rules/LaunchRulesEngine.swift
```diff
@@ -283,33 +283,33 @@ public class LaunchRulesEngine {
...
-                        return replaceToken(for: val as! String, data: data)
-                    } else {
-                        return val
-                    }
-                })
-                mutableDict[key] = tokenReplacedArr
-            default:
-                break
-            }
+            guard let value = value else { break }
```
praveek: We should not exit the loop when encountering null value. Can you change the logic so that it has same behaviour as before
```
     if (let value = value) {
          mutableDict[key] = replaceToken(in: value, data: data)
      }
```

praveek (approved):

PR #1001, opened 2024-02-15
[MOB-20427] - don't show fullscreenmessage with empty html

sbenedicadb:
## Types of changes

## Checklist:


AEPServices/Sources/ui/fullscreen/FullscreenMessage.swift
```diff
@@ -345,6 +345,14 @@
         }
 
         private func displayWithAnimation(webView: WKWebView) {
+            // MOB-20427 - only display the WKWebView if we have html to show
```
cdhoffmann, line 348: Is there a reason we can't do this check earlier at the beginning of the `handleShouldShow`? That way you don't inform the delegates if there is no HTML to begin with?

cdhoffmann (approved):

AEPServices/Sources/ui/fullscreen/FullscreenMessage.swift
cdhoffmann: > Is there a reason we can't do this check earlier at the beginning of the `handleShouldShow`? That way you don't inform the delegates if there is no HTML to begin with?
From Steve internally:
1. they want holdout tracking to behave like normal - i.e. they want "show" events sent to ajo even for a holdout experience (which is why the html is empty)
2. if the change gets put elsewhere, someone could call displayWithAnimation in the future without making the check for empty html.  this is the safer place to put the check as it covers all current and future scenarios

sbenedicadb: two reasons the code is where it is:

1. we want holdout tracking to behave like normal message tracking - i.e. we want "show" events sent to ajo even for a holdout experience (which is why the html is empty to begin with)
2. if the change gets put elsewhere, someone could call displayWithAnimation in the future without making the check for empty html.  this is the safer place to put the check as it covers all current and future scenarios

cdhoffmann: > Is there a reason we can't do this check earlier at the beginning of the `handleShouldShow`? That way you don't inform the delegates if there is no HTML to begin with?
From Steve internally:
1. they want holdout tracking to behave like normal - i.e. they want "show" events sent to ajo even for a holdout experience (which is why the html is empty)
2. if the change gets put elsewhere, someone could call displayWithAnimation in the future without making the check for empty html.  this is the safer place to put the check as it covers all current and future scenarios

PR #1020, opened 2024-05-02
Lifecycle v2 enhancement (app upgrade) 

yangyansong-adbe:
Port changes from the Android side: https://github.com/adobe/aepsdk-core-android/pull/650
The base branch will be updated after we figure out the release version.

### DIFF
--- a/AEPLifecycle/Sources/LifecycleV2/LifecycleV2.swift
+++ b/AEPLifecycle/Sources/LifecycleV2/LifecycleV2.swift
@@ -146,8 +146,8 @@ class LifecycleV2 {
 
     /// - Returns: Bool indicating whether the app has been upgraded
     private func isUpgrade() -> Bool {
-        if let currentAppVersion = systemInfoService.getApplicationBuildNumber(),
-           let previousAppVersion = dataStore.getString(key: LifecycleV2Constants.DataStoreKeys.LAST_APP_VERSION) {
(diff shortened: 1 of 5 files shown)


cdhoffmann (changes requested):
Looks good but have a couple of questions/comments:
1. Can we write a test to check that isUpgrade is behaving as expected?
2. Is there a reason we decided not to add this function as part of the SystemInfoService?

yangyansong-adbe:
> Looks good but have a couple of questions/comments:
> 
> 1. Can we write a test to check that isUpgrade is behaving as expected?
> 2. Is there a reason we decided not to add this function as part of the SystemInfoService?

1. This [functional test](https://github.com/adobe/aepsdk-core-ios/blob/main/AEPLifecycle/Tests/FunctionalTests/LifecycleV2FunctionalTests.swift#L165) covered the upgrade use case. Also, the changes are inside the private functions, it's hard to cover them in unit tests. 
2. It's a good idea to move it to `SystemInfoService` if other extensions need to call this function as well. Currently, this version string is only used in the Lifecycle extension, so I would recommend not moving it.

cdhoffmann (approved):

praveek:
Can you add the following tests 
1) In LifecycleV2Extension changing applicationVersionNumber also triggers an upgrade event
2) In LifecycleExtension changing applicationVersionNumber does not count as upgrade

praveek (approved):

PR #1022, opened 2024-05-13
Identity don't dispatch Analytics push event on privacy change

kevinlind:
## Description

Updates the Identity extension to not dispatch the `AnalyticsForIdentityRequest` event when privacy status is changed to optedout.

The Identity extension is expected to dispatch the `AnalyticsForIdentityRequest` event when the push identifier changes or when push tracking is denied the first time. This event tells Analytics to dispatch an ADBINTERNAL:Push event with `a.push.optin` value of either `True` or `False`. The issue is in some cases changing the privacy status to optedout would trigger a dispatch of the `AnalyticsForIdentityRequest` event even though `setPushIdentifier` was never called.

This fix changes how Identity handles processing of an privacy status change to optedout by resetting the `PushIdManager` flags instead of calling `updatePushId` which would dispatch the event to Analytics. 

## Types of changes

## Checklist:

### DIFF
--- a/AEPIdentity/Sources/IdentityState.swift
+++ b/AEPIdentity/Sources/IdentityState.swift
@@ -269,14 +269,7 @@ class IdentityState {
         identityProperties.privacyStatus = newPrivacyStatus
 
         if newPrivacyStatus == .optedOut {
-            identityProperties.ecid = nil
-            identityProperties.advertisingIdentifier = nil
-            identityProperties.blob = nil
-            identityProperties.locationHint = nil
-            identityProperties.customerIds?.removeAll()
-            identityProperties.isAidSynced = false
-            identityProperties.pushIdentifier = nil
-            pushIdManager.updatePushId(pushId: nil)
+            clearIdentifiers()
             identityProperties.saveToPersistence()
             createSharedState(identityProperties.toEventData(), event)
         } else if identityProperties.ecid == nil {
@@ -337,15 +330,7 @@ class IdentityState {
     func resetIdentifiers(event: Event,
                           createSharedState: ([String: Any], Event) -> Void) {
         guard identityProperties.privacyStatus != .optedOut else { return }
-        // clear the properties
-        identityProperties.ecid = nil
-        identityProperties.advertisingIdentifier = nil
-        identityProperties.blob = nil
-        identityProperties.locationHint = nil
-        identityProperties.customerIds?.removeAll()
-        identityProperties.isAidSynced = false
-        identityProperties.pushIdentifier = nil
-        pushIdManager.resetPersistedFlags()
+        clearIdentifiers()
         hitQueue.clear() // clear hit queue
 
         // do a force sync to generate ECID, then save the properties to persistence.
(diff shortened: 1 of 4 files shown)


AEPIdentity/Sources/IdentityState.swift
```diff
@@ -269,14 +269,7 @@ class IdentityState {
...
 
         if newPrivacyStatus == .optedOut {
-            identityProperties.ecid = nil
-            identityProperties.advertisingIdentifier = nil
-            identityProperties.blob = nil
-            identityProperties.locationHint = nil
-            identityProperties.customerIds?.removeAll()
-            identityProperties.isAidSynced = false
-            identityProperties.pushIdentifier = nil
-            pushIdManager.updatePushId(pushId: nil)
```
kevinlind, line 279: This call to `pushIdManager.updatePushId()` is replaced with `pushIdManager.resetPersistedFlags()` in the function `clearIdentifiers()`.

AEPIdentity/Sources/PushIDManager.swift
```diff
@@ -60,12 +60,13 @@ struct PushIDManager: PushIDManageable {
         properties.pushIdentifier = pushId
         properties.saveToPersistence()
 
-        if pushId?.isEmpty ?? true, !pushEnabled {
+        let isPushEnabled = pushEnabled
```
emdobrin, line 63: 👍 good catch

AEPIdentity/Tests/IdentityFunctionalTests.swift
```diff
@@ -467,4 +467,78 @@ class IdentityFunctionalTests: XCTestCase {
...
+        identity.state?.lastValidConfig = [IdentityConstants.Configuration.GLOBAL_CONFIG_PRIVACY: PrivacyStatus.optedIn.rawValue]
+        
+        let encodedPushId = "746573742D707573682D6964" // encoded value of string "test-push-id"
+                
+        let data = ["pushidentifier": encodedPushId]
+        let event = Event(name: "Set Push Identifier", type: EventType.genericIdentity, source: EventSource.requestContent, data: data)
+        
+        mockRuntime.simulateComingEvent(event: event)
+        
+        let dispatchedEvent = mockRuntime.dispatchedEvents.first
```
emdobrin, line 485: for these tests let's assert dispatchEvents count also

AEPIdentity/Tests/IdentityFunctionalTests.swift
```diff
@@ -467,4 +467,78 @@ class IdentityFunctionalTests: XCTestCase {
...
+
+        XCTAssertEqual(EventType.analytics, dispatchedEvent?.type)
+        XCTAssertEqual(EventSource.requestContent, dispatchedEvent?.source)
+        XCTAssertEqual("Push", dispatchedEvent?.data?["action"] as? String)
+        XCTAssertTrue((dispatchedEvent?.data?["trackinternal"] as? Bool) ?? false)
+        let contextData = dispatchedEvent?.data?["contextdata"] as? [String: Any]
+        XCTAssertEqual("True", contextData?["a.push.optin"] as? String)
```
emdobrin, line 492: - could use json comparison for easier reading
- can we check the persisted value for push token as well

cdhoffmann: @kevinlind https://github.com/adobe/aepsdk-core-ios/pull/1002

We had to include the files as source because AEPTestUtils depends on Core and caused a circular dependency compilation error. We have a task for a long term solution in the future.

emdobrin (approved):

praveek (approved):

PR #1026, opened 2024-05-16
Retry hit processor requests failing with recoverable URL errors

praveek:
Currently, the hit processor drops a hit even when the network request fails due to a recoverable error. Fix the processing logic to retry when the network failure occurs due to one of the `NetworkServiceConstants.RECOVERABLE_URL_ERROR_CODES`.

### DIFF
--- a/AEPIdentity/Sources/IdentityHitProcessor.swift
+++ b/AEPIdentity/Sources/IdentityHitProcessor.swift
@@ -65,7 +65,7 @@ class IdentityHitProcessor: HitProcessing {
             // retry this hit later
             Log.debug(label: "\(LOG_TAG):\(#function)", "Retrying Identity hit, request with url \(hit.url.absoluteString) failed with error \(connection.error?.localizedDescription ?? "") and recoverable status code \(connection.responseCode ?? -1)")
             completion(false)
-        } else if let error = connection.error as? URLError, NetworkServiceConstants.RECOVERABLE_URL_ERROR_CODES.contains(error.code) {
+        } else if let error = connection.error as? URLError, error.isRecoverable {
             // retry this hit later as the request failed with a recoverable transport error
             Log.debug(label: "\(LOG_TAG):\(#function)", "Retrying Identity hit, request with url \(urlString) failed with error \(connection.error?.localizedDescription ?? "") and recoverable status code \(connection.responseCode ?? -1)")
             completion(false)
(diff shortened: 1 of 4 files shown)


AEPIdentity/Sources/IdentityHitProcessor.swift
```diff
@@ -65,6 +65,10 @@ class IdentityHitProcessor: HitProcessing {
             // retry this hit later
             Log.debug(label: "\(LOG_TAG):\(#function)", "Retrying Identity hit, request with url \(hit.url.absoluteString) failed with error \(connection.error?.localizedDescription ?? "") and recoverable status code \(connection.responseCode ?? -1)")
             completion(false)
+        } else if let error = connection.error as? URLError, NetworkServiceConstants.RECOVERABLE_URL_ERROR_CODES.contains(error.code) {
```
addb, line 68: You can use the extension method that got added #1027 
```
if let error = connection.error as? URLError, urlError.isRecoverable)
```

AEPSignal/Sources/SignalHitProcessor.swift
```diff
@@ -70,6 +70,10 @@ class SignalHitProcessor: HitProcessing {
             // retry this hit later
             Log.debug(label: LOG_TAG, "Signal request failed with recoverable error \(connection.error?.localizedDescription ?? "") and status code \(connection.responseCode ?? -1). Will retry sending the request later: \(hit.url.absoluteString)")
             completion(false)
+        } else if let error = connection.error as? URLError, NetworkServiceConstants.RECOVERABLE_URL_ERROR_CODES.contains(error.code) {
```
addb, line 73: You can use the extension method that got added #1027 
```
if let error = connection.error as? URLError, error.isRecoverable)
```

AEPIdentity/Tests/IdentityHitProcessorTests.swift
```diff
@@ -73,34 +73,40 @@ class IdentityHitProcessorTests: XCTestCase {
...
-        mockNetworkService?.setMockResponse(url: expectedUrl, httpMethod: .get, responseConnection: testConnection)
-
-        let entity = DataEntity(uniqueIdentifier: "test-uuid", timestamp: Date(), data: try! JSONEncoder().encode(hit))
-
-        // test
-        hitProcessor.processHit(entity: entity) { success in
-            XCTAssertFalse(success)
-            expectation.fulfill()
+        NetworkServiceConstants.RECOVERABLE_ERROR_CODES.forEach { error in
+            let expectation = XCTestExpectation(description: "Callback should be invoked with true signaling this hit should not be retried")
```
addb, line 85: "Callback should be invoked with false signaling this hit should be retried"

addb (approved):

PR #1046, opened 2024-06-07
More stability fixes

praveek:
- Prevent concurrent access exceptions in ConfigurationState.
- Update `ConfigurationState` tests to use the validation method from test utils.
- Pass weak references to network response handler.

### DIFF
--- a/AEPCore/Sources/configuration/ConfigurationState.swift
+++ b/AEPCore/Sources/configuration/ConfigurationState.swift
@@ -17,22 +17,20 @@ import Foundation
 class ConfigurationState {
     private static let CONFIGURATION_TTL = TimeInterval(15)
 
-    let dataStore: NamedCollectionDataStore
-    let appIdManager: LaunchIDManager
-    let configDownloader: ConfigurationDownloadable
+    private let queue = DispatchQueue(label: "com.adobe.configurationState")
+    private let dataStore: NamedCollectionDataStore
+    private let appIdManager: LaunchIDManager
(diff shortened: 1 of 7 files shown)


AEPCore/Sources/configuration/ConfigurationState.swift
```diff
@@ -57,104 +55,81 @@ class ConfigurationState {
...
-
-        updateWith(newConfig: config)
     }
 
-    /// Merges the current configuration to `newConfig` then applies programmatic configuration on top
+    /// Updates the current configuration to `newConfig` then applies programmatic configuration on top
     /// - Parameter newConfig: The new configuration
     func updateWith(newConfig: [String: Any]) {
-        unmergedConfiguration = newConfig
-        currentConfiguration.merge(newConfig) { _, updated in updated }
```
cdhoffmann, line 81: why are we changing this logic?

praveek: None of our APIs support merging the configuration with a provided dictionary. We were only using this functionality within loadInitialConfig(), which simply sets the configuration initially. Even if we had supported this, the current logic would still be incorrect. The unmergedConfiguration should also be merged with the provided config.

cdhoffmann: Ahhhh makes sense. Thanks for the clarification.

cdhoffmann (approved):

PR #1050, opened 2024-06-20
Fix race condition in registerEventListener

praveek:
Currently, the EventHubPlaceHolder extension is automatically registered in the init method of EventHub. The EventHub.registerEventListener() method uses this placeholder instance to store all listeners. We are presently fetching the extensionContainer for the EventHub extension using the shared state name, which becomes available only after the extension is initialized asynchronously. This leads to a lot of flaky test failures. In real-world scenarios, the impact is minimal but could occur if the app attempts to register an event listener early in the app lifecycle.

To resolve this, we will pull the EventHub extension container using its type instead of the shared state name to ensure listeners are not missed during registration. Additionally, helper methods have been added to avoid duplicate logic inside EventHub.

### DIFF
--- a/AEPCore/Sources/eventhub/EventHub.swift
+++ b/AEPCore/Sources/eventhub/EventHub.swift
@@ -134,9 +134,9 @@ final class EventHub {
             }
 
             // Init the extension on a dedicated queue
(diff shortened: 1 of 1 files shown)


sbenedicadb (approved):

cdhoffmann (approved):
Good find!

PR #1058, opened 2024-07-23
Improve AEPTestUtils Network Service helper class multi thread handling and method docs

timkimadobe:
## Description

This PR updates the AEPTestUtils NetworkService helper classes to better handle multi thread access by using a class level queue that protects access to underlying data storage. Async await operations are handled outside of the queue so that the awaiting task doesn't block the actual recording of what they are trying to assert against, within the same serial queue.

It also updates the: 
1. Method documentation for clarity and consistency, also adding method docs for some methods that were missing them previously.
2. Ordering of methods so they are more closely grouped by functionality and dependent order

## Types of changes

## Checklist:

### DIFF
--- a/AEPServices/Mocks/PublicTestUtils/NetworkRequestHelper.swift
+++ b/AEPServices/Mocks/PublicTestUtils/NetworkRequestHelper.swift
@@ -21,6 +21,7 @@ import XCTest
 ///    - ``MockNetworkService``
 ///    - ``RealNetworkService``
 class NetworkRequestHelper {
+    private let LOG_TAG = "NetworkRequestHelper"
     /// A dispatch queue for synchronizing access to the internal state.
     private let queue = DispatchQueue(label: "com.adobe.networkrequesthelper.syncqueue")
 
@@ -58,7 +59,7 @@ class NetworkRequestHelper {
     ///
     /// - Parameter networkRequest: The ``NetworkRequest`` object representing the network request that was sent.
     func recordSentNetworkRequest(_ networkRequest: NetworkRequest) {
-        print("Received connectAsync to URL \(networkRequest.url.absoluteString) and HTTPMethod \(networkRequest.httpMethod.toString())")
+        Log.debug(label: LOG_TAG, "Received connectAsync to URL \(networkRequest.url.absoluteString) and HTTPMethod \(networkRequest.httpMethod.toString())")
 
         queue.async { [weak self] in
             guard let self = self else { return }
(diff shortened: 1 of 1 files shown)


AEPServices/Mocks/PublicTestUtils/MockNetworkService.swift
```diff
@@ -108,58 +118,30 @@ public class MockNetworkService: Networking {
...
     ///
-    /// Use this method after calling `setExpectationForNetworkRequest(url:httpMethod:expectedCount:file:line:)` to wait for expected requests.
+    /// This method waits for the specified network request to be fulfilled within the provided timeout interval,
+    /// ensuring that any expected requests are received. If no expectation exists for the provided request,
+    /// the method will perform a regular wait for the specified timeout duration.
     ///
     /// - Parameters:
-    ///   - url: The URL `String` of the `NetworkRequest` to get.
-    ///   - httpMethod: The HTTP method of the `NetworkRequest` to get.
+    ///   - url: The URL ``String`` of the ``NetworkRequest`` to get.
```
addb, line 139: code formatting is being applied twice for String and Network request and similar to other function doc updates that have been done.

praveek (approved):

AEPServices/Mocks/PublicTestUtils/MockNetworkService.swift
addb: Got it. Nothing to be done.
https://stackoverflow.com/questions/27715933/how-to-use-swift-documentation-comments

AEPServices/Mocks/PublicTestUtils/MockNetworkService.swift
```diff
@@ -108,58 +118,30 @@ public class MockNetworkService: Networking {
...
+    ///   - httpMethod: The HTTP method of the ``NetworkRequest`` to get.
     ///   - timeout: The time interval to wait for network requests before timing out. Defaults to ``TestConstants/Defaults/WAIT_NETWORK_REQUEST_TIMEOUT``.
     ///   - file: The file from which the method is called, used for localized assertion failures.
     ///   - line: The line from which the method is called, used for localized assertion failures.
-    /// - Returns: An array of `NetworkRequest`s that match the provided `url` and `httpMethod`. Returns an empty array if no matching requests were dispatched.
+    /// - Returns: An array of ``NetworkRequest``s that match the provided ``url`` and ``httpMethod``. Returns an empty array if no matching requests were dispatched.
```
addb, line 144: How much is the usage of this function, should we deprecate this in favor of the `getNetworkRequestWith(url: URL, ...)` ?

addb: @timkimadobe 
I believe it's easier setup to use string and have the helper convert to URL. I wanted see if we can have only one helper with either string or URL. Do you know if we have lot of usages with url:URL signature as well?

addb (approved):

PR #1059, opened 2024-07-23
Update JSON comparison APIs

timkimadobe:
> [!IMPORTANT]
> Test cases have not been included in this PR in order to keep size manageable; test cases validating this implementation will be opened in follow up PRs.

## Description

This PR updates the JSON comparison APIs with three main changes that bring the iOS APIs up to date with the latest functionality added to the Android version of the APIs:
1. **Breaking change:** Removes the deprecated APIs that used different path interpretation logic (and associated methods).
2. New path options:
    1. ElementCount - the number of leaf nodes must be equal to the set expectation
    2. ValueNotEqual - the value at the node must not be equal to what is in expected

### 1. Breaking change: Removes the deprecated APIs that used different path interpretation logic (and associated methods).
`XCTestCase+AnyCodableAsserts.swift`
Removed deprecated APIs:
`assertTypeMatch(expected:actual:exactMatchPaths:file:line:)`
`assertExactMatch(expected:actual:typeMatchPaths:file:line:)`
These methods only allowed for a very inflexible switch between value type/exact validation modes, and also baked in array AnyOrder path string interpretation into the method logic.

`NodeConfig.swift`
Removed `extractRegexCaptureGroups` and `getArrayIndexAndAnyOrder`. 
* `getArrayIndexAndAnyOrder` was used in the deprecated APIs to interpret path strings like `[*1]` and `[*]` as the AnyOrder path option. Instead, the current APIs require you to specify this using the path option `AnyOrder`. The motivation was to move away from using the `*` character as AnyOrder, and instead as a true wildcard name for dictionary keys and array indices.

### 2i. ElementCount - the number of leaf nodes must be equal to the set expectation
Parity with Android (motivation also explained in PR): https://github.com/adobe/aepsdk-testutils-android/pull/25

### 2ii. ValueNotEqual - the value at the node must not be equal to what is in expected
Parity with Android: https://github.com/adobe/aepsdk-testutils-android/pull/24

## Types of changes

## Checklist:

### DIFF
--- a/AEPServices/Mocks/PublicTestUtils/XCTestCase+AnyCodableAsserts.swift
+++ b/AEPServices/Mocks/PublicTestUtils/XCTestCase+AnyCodableAsserts.swift
@@ -18,7 +18,7 @@ import XCTest
 ///
 /// `ValidationResult` is used by the `validateActual` recursive methods to return the outcome of their validation
 /// and to track the number of leaf nodes encountered during a depth-first traversal.
-struct ValidationResult {
+private struct ValidationResult {
     /// A Boolean value indicating whether the validation was successful.
     var isValid: Bool
 
@@ -1023,7 +1023,7 @@ public extension AnyCodableAsserts where Self: XCTestCase {
     ///   - line: The line from which the method is called, used for localized assertion failures.
     ///
     /// - Returns: A `Bool` indicating whether the `actual` value is valid based on the `nodeTree` configuration.
-    internal func validateActual(
+    private func validateActual(
         actual: AnyCodable?,
         keyPath: [Any] = [],
         nodeTree: NodeConfig,


AEPServices/Mocks/PublicTestUtils/XCTestCase+AnyCodableAsserts.swift
```diff
@@ -14,6 +14,11 @@ import AEPServices
 import Foundation
 import XCTest
 
+struct ValidationResult {
+    var isValid: Bool
+    var elementCount: Int
```
praveek, line 19: What does `elementCount` mean? Can you add some comments around it?

cacheung (approved):

AEPServices/Mocks/PublicTestUtils/XCTestCase+AnyCodableAsserts.swift
praveek: Can you mark it as private as it’s used only within this file?

Issue #1064, opened 2024-08-15
Investigate EventHub and ExtensionContainer shutdown race condition


timkimadobe:
### Task description

Currently in `IdentityIntegrationTests.swift`, test cases rely on the `tearDown` operation to unregister all extensions, shut down the EventHub and reset the EventHub. However, there are some cases especially when rerunning many iterations of a test case without rebuilding where events can carry over across test cases. 

The current hypothesis is that the chain of events is:
1. Identity extension dispatching response event (is the async operation)
2. EventHub[run1] shutdown + reset
3. EventHub[run2] starts up
4. EventHub[run2] receives event from (1.)

When the correct case should be:
1. Identity extension dispatching response event (is the async operation)
2. EventHub[run1] receives event from (1.)
3. EventHub[run1] shutdown + reset
4. EventHub[run2] starts up

I tried catching this case in the logs using `print` and `os_log` to trace:
`EventHub`
1. `init()`
2. `dispatch(event:)`
3. `shutdown()`

`public extension Extension`
1. `dispatch(event:)`

However, both methods appear to be too heavy and affect the actual execution speed either:
1. Causing regular test setup timeout assertions to fail
2. Creating new types of timeout errors (ex: where response events that would have been dispatched in time are not)

If test flakiness due to this issue needs to be addressed and resolving this architectural interaction is too costly, then another workaround could be to use event based expectations and assertions (which await for those expectations) to properly wait for all events from a given test case to process before proceeding with teardown operations and the next test case.

PR #1065, opened 2024-08-16
Use instanced EventHub with ExtensionContainer

timkimadobe:
## Description

This PR updated the ExtenionContainer class to use an instance of EventHub instead of a shared global reference.

This is to avoid a relatively common test scenario where EventHub is started and shutdown rapidly between test cases, and extension containers have pending events that will get sent to the new EventHub instance; this can affect test case state, shared states, processing times, etc.

See: https://github.com/adobe/aepsdk-core-ios/issues/1064

## Types of changes

## Checklist:


praveek (approved):

sbenedicadb (approved):
i think the change is good - i'm always hesitant to make a change like this because even though it's quite small, it is in the hottest piece of code in our whole sdk - let's make sure we do some thorough regression testing across the board on this one before releasing it

PR #1073, opened 2024-09-18
Update EventHub with Multi-Instance Support

praveek:
**EventHub**
 - Updated EventHub to support multiple instances.
 - Updated labels for DispatchQueue and EventQueue to include instance information. Labels for `ThreadSafeArray` and `ThreadSafeDictionaries` were not updated, as they would not provide useful information and would add unnecessary clutter. 
 - `EventHub.shared` was not cleaned up in this PR; it will be addressed in a separate PR, which will involve significant updates to our tests. 
 
**ExtensionRuntime**
- Added a new API in to expose `ExtensionServiceProvider` to extensions.

**ExtensionContainer**
- Cleanup dependencies and implemented the new `ExtensionRuntime` API.
- Moved `EventPreprocessor` registration to `ExtensionContainer`. This API should ideally be exposed through `ExtensionRuntime` to avoid such workarounds. This issue is outside the scope of multi-instance implementation and will be addressed separately. 

**EventHistory**
- Updated `EventHistory` to support multiple instances.
- Cleaned up current tests 
- Added new tests to verify multi-instance functionality.

**Other**
- Add the `EventHubProvider` class. Tests and integration for this class will be included in a follow-up PR as part of the Core API implementation.

### DIFF
--- a/AEPCore/Sources/eventhub/EventHub.swift
+++ b/AEPCore/Sources/eventhub/EventHub.swift
@@ -22,26 +22,36 @@ public typealias EventPreprocessor = (Event) -> Event
 /// Responsible for delivering events to listeners and maintaining registered extension's lifecycle.
 final class EventHub {
     private let LOG_TAG = "EventHub"
-    private let eventHubQueue = DispatchQueue(label: "com.adobe.eventHub.queue")
+    private let identifier: SDKInstanceIdentifier
+    private let logger: Logger
+    private let extensionServiceProvider: ExtensionServiceProvider
+    private let eventHubQueue: DispatchQueue
+    private let eventQueue: OperationOrderer<Event>
     private var registeredExtensions = ThreadSafeDictionary<String, ExtensionContainer>(identifier: "com.adobe.eventHub.registeredExtensions.queue")
     private let eventNumberMap = ThreadSafeDictionary<UUID, Int>(identifier: "com.adobe.eventHub.eventNumber.queue")
     private let responseEventListeners = ThreadSafeArray<EventListenerContainer>(identifier: "com.adobe.eventHub.response.queue")
     private var eventNumberCounter = AtomicCounter()
-    private let eventQueue = OperationOrderer<Event>("EventHub")
     private var preprocessors = ThreadSafeArray<EventPreprocessor>(identifier: "com.adobe.eventHub.preprocessors.queue")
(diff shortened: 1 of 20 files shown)


AEPCore/Sources/eventhub/EventHub.swift
```diff
@@ -394,8 +414,8 @@ final class EventHub {
         self.eventQueue.add(event)
         // Hot path, avoid unnecessary string converstion of event
         if Log.logFilter >= .trace {
```
kevinlind, line 416: Should we consider adding `logFilter` (or `logLevel`) to the `Logger` protocol so we don't need to access the public Log class?

kevinlind (approved):
Looks great! I've added two minor questions, but they can be addressed in another PR if needed.

AEPCore/Sources/eventhub/EventHub.swift
```diff
@@ -22,26 +22,36 @@ public typealias EventPreprocessor = (Event) -> Event
...
-    private let eventHubQueue = DispatchQueue(label: "com.adobe.eventHub.queue")
+    private let identifier: SDKInstanceIdentifier
+    private let logger: Logger
+    private let extensionServiceProvider: ExtensionServiceProvider
+    private let eventHubQueue: DispatchQueue
+    private let eventQueue: OperationOrderer<Event>
     private var registeredExtensions = ThreadSafeDictionary<String, ExtensionContainer>(identifier: "com.adobe.eventHub.registeredExtensions.queue")
     private let eventNumberMap = ThreadSafeDictionary<UUID, Int>(identifier: "com.adobe.eventHub.eventNumber.queue")
     private let responseEventListeners = ThreadSafeArray<EventListenerContainer>(identifier: "com.adobe.eventHub.response.queue")
     private var eventNumberCounter = AtomicCounter()
```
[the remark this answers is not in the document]
praveek: Having each EventHub maintain its own eventNumberCounter will simplify debugging. Otherwise, event numbers and shared state versions would become unpredictable, even for related events.

AEPCore/Sources/eventhub/EventHub.swift
praveek: That makes sense. Adding a `logFilter` getter will make it easier to control the logs for each SDKInstanceLogger in the future. I’ll include this update in my next PR

PR #1080, opened 2024-09-27
MobileCore API changes for multi instance support

praveek:
## Types of changes

## Checklist:

### DIFF
--- a/AEPCore/Sources/core/MobileCore+MultiInstance.swift
+++ b/AEPCore/Sources/core/MobileCore+MultiInstance.swift
@@ -1,0 +1,59 @@
+/*
+ Copyright 2024 Adobe. All rights reserved.
+ This file is licensed to you under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License. You may obtain a copy
+ of the License at http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software distributed under
+ the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ OF ANY KIND, either express or implied. See the License for the specific language
+ governing permissions and limitations under the License.
+*/
+
+import AEPServices
+
+/// An extension for `MobileCore` that provides methods to create and manage instances of the SDK.
+/// This includes the ability to create named instances and retrieve the API for specific instances.
+@objc
+public extension MobileCore {
+
+    /// Initializes a named instance of the SDK with the specified name.
+    /// This method must be called first before calling any other API for the given instance.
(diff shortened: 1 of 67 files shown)


AEPCore/Sources/core/MobileCoreAPI.swift
```diff
@@ -0,0 +1,413 @@
...
+    }
+    
+    private func getEventHub(_ debugInfo: String) -> EventHub? {
+        guard let sdkInstanceIdentifier = sdkInstanceIdentifier else {
+            Log.warning(label: logTag, "MobileCoreAPI::\(debugInfo) - SDK instance name \(instanceIdentifier) is invalid.")
+            return nil
+        }
+        
+        guard let eventHub = eventHubProvider.getEventHub(for: sdkInstanceIdentifier) else {
+            Log.warning(label: logTag, "MobileCoreAPI::\(debugInfo) - SDK instance \(instanceIdentifier) has not been initialized.")
```
kevinlind, line 41: nit: reword to "could not be initialized" as "has not been" seems to imply that it still could be initialized in the future.

AEPCore/Sources/core/MobileCore+MultiInstance.swift
```diff
@@ -0,0 +1,59 @@
...
+    /// - Parameter name: The name for the SDK instance. The name must be valid; if it's invalid, instance creation will fail.
+    @objc(initializeInstance:)
+    static func initializeInstance(_ name: String) {
+        guard let identifier = SDKInstanceIdentifier(id: name) else {
+            Log.warning(label: LOG_TAG, "Unable to initialize as SDK instance name '\(name)' is invalid.")
+            return
+        }
+                
+        eventHubProvider.createEventHub(for: identifier)
+    }
```
kevinlind, line 31: I think we should return something here, such as a Bool, to indicate if the initialization failed or succeeded. There is no other API which returns something to the application which indicates if an SDK instance was created or not, so this could serve that purpose.

kevinlind (commented):
This looks great. I've included two minor comments.

kevinlind (approved):

AEPCore/Sources/core/MobileCoreAPI.swift
praveek: The initialization may occur later. The error is thrown at this point so the SDK can return a non-nil instance for `MobileCore.api(for: instance)`, simplifying SDK usage.

AEPCore/Sources/core/MobileCore+MultiInstance.swift
praveek: The failure only happens when the name is in an unsupported format, which is logged and can be caught during development.

PR #1083, opened 2024-10-02
Fix TestBase getSharedStateFor

timkimadobe:
## Description

This PR fixes the getSharedStateFor method in TestBase by updating its return type from `[AnyHashable: Any]?` to `SharedStateResult`.

The existing logic was silently failing to cast the actual value type of `SharedStateResult` (NSObject) into the `[AnyHashable: Any]?` and always returning `nil` whether a valid shared state existed or not.

Logic chain:
1. TestBase's getSharedStateFor dispatches an InstrumentedExtension shared state request event: https://github.com/adobe/aepsdk-testutils-ios/blob/5ec3b66f4c74ecc639a8471790628bd84925c59d/Sources/TestBase.swift#L193
2. InstrumentedExtension handles this event by calling: https://github.com/adobe/aepsdk-testutils-ios/blob/5ec3b66f4c74ecc639a8471790628bd84925c59d/Sources/InstrumentedExtension.swift#L92
3. `ExtensionRuntime`'s `getSharedState` returns a `SharedStateResult`: https://github.com/adobe/aepsdk-core-ios/blob/eab8353a2ccc2d3cc760c98f8e48e2dd9fc5168d/AEPCore/Sources/eventhub/ExtensionRuntime.swift#L68
4. `InstrumentedExtension` sets this value as the state: https://github.com/adobe/aepsdk-testutils-ios/blob/5ec3b66f4c74ecc639a8471790628bd84925c59d/Sources/InstrumentedExtension.swift#L93

So when using TestBase's `getSharedStateFor` method, the value of the `state` key is always a `SharedStateResult` anyways.

Currently there appears to only be a single usage of this API in iOS: 
NoConfigFunctionalTests.swift - testHandleExperienceEventRequest_withPendingConfigurationState_expectEventsQueueIsBlocked(): https://github.com/adobe/aepsdk-edge-ios/blob/726f7b158a80fea0b827ceca5ddc9e71932689bf/Tests/FunctionalTests/NoConfigFunctionalTests.swift#L50
* In this test case, it actually wants a nil value to verify that configuration is in a pending state (shared state will be nil), so the return value failing to cast causes this check to always pass as long as the getting of the shared state itself doesn't time out.

Proposing updating this method to be a passthrough to EventHub's `getSharedState(extensionName:event:barrier:resolution:sharedStateType:)` method, with the main difference being that the `SharedStateType` is hardcoded to `.standard` (as `SharedStateType` is not a public type).

This is a breaking change, but given that:
1. The existing behavior doesn't work as expected (always returns `nil`), and this behavior is strongly tied to the return type itself
2. There's only 1 usage
I believe the change is worth it

## Types of changes

## Checklist:

### DIFF
--- a/AEPCore/Mocks/PublicTestUtils/TestBase.swift
+++ b/AEPCore/Mocks/PublicTestUtils/TestBase.swift
@@ -17,13 +17,6 @@ import AEPServices
 @testable import AEPCore
 
 open class TestBase: XCTestCase {
-    /// Represents the `SharedState` types in ``EventHub`` that can be used by an extension to share data with other extensions and for rules execution.
-    /// This is a publicly accessible version of the ``SharedStateType`` used internally.
-    public enum SharedStateTypeTestHelper: String {
-        case standard = "standard"  // regular data, the key names and structure can be defined by each extension
-        case xdm = "XDM"  // data mapped on XDM mixins populated by an extension
-    }
-
     /// Use this property to execute code logic in the first run in this test class; this value changes to False after the parent tearDown is executed
     public private(set) static var isFirstRun: Bool = true
     /// Use this setting to enable debug mode logging in the `TestBase`
@@ -190,7 +183,7 @@ open class TestBase: XCTestCase {
         return InstrumentedExtension.receivedEvents[EventSpec(type: type, source: source)] ?? []
     }
 
-    /// Retrieves the `SharedState` for a specific extension.
+    /// Retrieves the standard `SharedState` for a specific extension (as opposed to XDM).
     /// This method fetches the shared state for a given extension name, optionally based on a specific event.
     /// The shared state can be resolved according to the specified resolution and shared state type.
     /// - Parameters:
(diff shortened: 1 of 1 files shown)


kevinlind (approved):

AEPCore/Mocks/PublicTestUtils/TestBase.swift
```diff
@@ -183,24 +183,31 @@ open class TestBase: XCTestCase {
...
 
-        var returnedState: [AnyHashable: Any]?
+        var returnedState: SharedStateResult?
 
         let expectation = XCTestExpectation(description: "Shared state data returned")
         MobileCore.dispatch(event: event, responseCallback: { event in
 
             if let eventData = event?.data {
-                returnedState = eventData["state"] as? [AnyHashable: Any]
+                returnedState = eventData[TestConstants.EventDataKey.STATE] as? SharedStateResult
```
praveek, line 210: We can change the implementation later but I prefer calling `EventHub.shared.getSharedState` directly as it’s less error-prone and doesn’t risk timing out.

praveek (approved):

AEPCore/Mocks/PublicTestUtils/TestBase.swift
```diff
@@ -183,44 +183,23 @@ open class TestBase: XCTestCase {
...
-
-        wait(for: [expectation], timeout: timeout)
-        return returnedState
+    ///   - extensionName: An extension name whose `SharedState` will be returned
+    ///   - event: If not nil, will retrieve the `SharedState` that corresponds with this event's version or latest if not yet versioned. If event is nil will return the latest `SharedState`
+    ///   - barrier: If true, the `EventHub` will only return `.set` if `extensionName` has moved past `event`
+    ///   - resolution: The `SharedStateResolution` to determine how to resolve the shared state
+    ///   - sharedStateType: The type of shared state to be read from, if not provided defaults to `.standard`
+    /// - Returns: The `SharedState` data and status for the extension with `extensionName`
+    public func getSharedStateFor(extensionName: String, event: Event?, barrier: Bool = true, resolution: SharedStateResolution = .any, sharedStateType: SharedStateType = .standard) -> SharedStateResult? {
```
praveek, line 194: Can you set the default value of event to nil, so it returns the latest shared state value if no event is provided?

praveek: Since SharedStateType is internal to Core, can you remove the parameter and default this API to the standard shared state? If needed, we can add a separate API for XDM shared state later.

praveek (approved):

PR #1084, opened 2024-10-02
TestUtils cleanup and updates

timkimadobe:
## Description

This PR does some cleanup of TestUtils related classes and usages: 
1. Removes the `isFirstRun` Bool from TestBase. 
    * If first run setup behavior is required, we can use the class level setup/teardown methods: https://developer.apple.com/documentation/xctest/xctestcase/set_up_and_tear_down_state_in_your_tests#3909114
    * Based on search through repos, it doesn't look like this Bool is used today and should be safe to remove
2. Removes `static` from `TestBase` property `debugEnabled` and rename to `loggingEnabled`
    * Rename motivation is to make the intent of the property more clear. Its only purpose is to control whether print logs are sent out or not; it does not affect any debugging behavior or business logic
    * `static` to non-static is also already a breaking change anyways, so bundling these changes together is probably the most optimal timing
3. Add deprecation warning to InstrumentedExtension, as the intent is to replace this (hacky) class with a wildcard listener implementation instead
4. Replace usages of EventHub.reset() with MobileCore.resetSDK() throughout project tests

## Types of changes

## Checklist:

### DIFF
--- a/AEPCore/Mocks/PublicTestUtils/InstrumentedExtension.swift
+++ b/AEPCore/Mocks/PublicTestUtils/InstrumentedExtension.swift
@@ -16,6 +16,7 @@ import AEPCore
 import AEPServices
 
 /// Instrumented extension that registers a wildcard listener for intercepting events in current session. Use it along with `TestBase`
+@available(*, deprecated, message: "InstrumentedExtension is deprecated and will be replaced by a new wildcard listener class soon. Please avoid using InstrumentedExtension in your tests.")
 public class InstrumentedExtension: NSObject, Extension {
     private static let logTag = "InstrumentedExtension"
     public var name = "com.adobe.InstrumentedExtension"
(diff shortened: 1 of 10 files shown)


AEPCore/Mocks/PublicTestUtils/TestBase.swift
```diff
@@ -17,8 +17,6 @@ import AEPServices
 @testable import AEPCore
 
 open class TestBase: XCTestCase {
-    /// Use this property to execute code logic in the first run in this test class; this value changes to False after the parent tearDown is executed
-    public private(set) static var isFirstRun: Bool = true
     /// Use this setting to enable debug mode logging in the `TestBase`
     public static var debugEnabled = false
```
praveek, line 21: Could you also check and remove this boolean? If it’s being used somewhere, please make it non-static.

AEPCore/Mocks/PublicTestUtils/InstrumentedExtension.swift
```diff
@@ -16,6 +16,7 @@ import AEPCore
 import AEPServices
 
 /// Instrumented extension that registers a wildcard listener for intercepting events in current session. Use it along with `TestBase`
+@available(*, deprecated, message: "Will be replaced by a new wildcard listener class soon. Please avoid using InstrumentedExtension directly in your tests.")
```
praveek, line 19: A new class for capturing events using MobileCore.registerEventListener will be added in the next version. Please avoid using InstrumentedExtension directly in your tests.

praveek (approved):

PR #1102, opened 2024-11-21
[WIP] Bugfix : Reframing IAM during iPad splitScreen

PravinPK:
Fixed a FullScreenMessage bug that does not redraw when the application enters split screen mode.

Used NSKeyValueObservation to observer changes to the Application's key UIWindow's frame to detect device rotation and split screen size change.

The fix is tested on following Apps
- Application based on Scene Delegate
- Application based on UIAppDelegate lifecycle 
- Complete SwiftUI Application

The fix was tested on various orientation and status bar settings
Screenshot 2024-11-21 at 3 47 27 PM

The fix is tested on the following devices
- iPhone 18 Simulator
- iPhone 18 device
- iPad 18 Simulator
- iPad 14 device

Simulator Screen Recording - iPad Pro - 2024-11-21 at 14 44 13

### DIFF
--- a/AEPServices/Sources/ui/fullscreen/FullscreenMessage.swift
+++ b/AEPServices/Sources/ui/fullscreen/FullscreenMessage.swift
@@ -48,7 +48,7 @@
         var messagingDelegate: MessagingDelegate? {
             return ServiceProvider.shared.messagingDelegate
         }
-        
+
         /// Observes key window frame changes to trigger frame updates for Fullscreen Message.
         private var windowFrameObserver: NSKeyValueObservation?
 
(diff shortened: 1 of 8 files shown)


AEPServices/Sources/ui/fullscreen/FullscreenMessage.swift
```diff
@@ -120,12 +124,16 @@
 
             DispatchQueue.main.async {
 
-                // add observer to handle device rotation
                 if !self.observerSet {
-                    NotificationCenter.default.addObserver(self,
-                                                           selector: #selector(self.handleDeviceRotation(notification:)),
-                                                           name: UIDevice.orientationDidChangeNotification,
-                                                           object: nil)
```
PravinPK, line 128: No need to for listen for notification : UIDevice.orientationDidChangeNotification
Because windowFrameObserver will also gets triggered when the device orientation changes

PR #1113, opened 2025-02-07
Simplified initialization for AEP SDK

kevinlind:
Add simplified initialization for AEP SDK.

Added MobileCore.initialize(...) APIs
Added MobileCoreInitializer class to handle initialization
Added InitOptions class
The new API automatically registers bundled extensions, sets up lifecycle tracking, and configures the SDK.
Added ClassFinder utility in AEPServices which finds all iOS registered classes based on a given Protocol


AEPCore/Sources/core/MobileCoreInitializer.swift
```diff
@@ -0,0 +1,108 @@
...
+            if #available(iOS 13.0, tvOS 13.0, *) {
+                NotificationCenter.default.addObserver(forName: UIScene.willEnterForegroundNotification, object: nil, queue: nil) { _ in
+                    MobileCore.lifecycleStart(additionalContextData: additionalContextData)
+                }
+                NotificationCenter.default.addObserver(forName: UIScene.didEnterBackgroundNotification, object: nil, queue: nil) { _ in
+                    MobileCore.lifecyclePause()
+                }
+            }
+        } else {
+            if UIApplication.shared.applicationState != .background {
```
praveek, line 97: This should be called from the UI thread.

AEPCore/Sources/core/MobileCoreInitializer.swift
```diff
@@ -0,0 +1,108 @@
...
+            MobileCore.setAppGroup(options.appGroup)
+        }
+
+        // Register Extensions, call configureWithAppId from callback
+        DispatchQueue.global().async {
+            let classList = self.classFinder(Extension.self)
+            let filteredClassList = classList.filter { $0 !== AEPCore.EventHubPlaceholderExtension.self && $0 !== AEPCore.Configuration.self }.compactMap { $0 as? NSObject.Type }
+            MobileCore.registerExtensions(filteredClassList) {
+
+                if let appId = options.appId {
```
praveek, line 56: Can you modify the logic to be similar to Android, ensuring that we call these outside the callback after setting appGroup?

AEPCore/Sources/core/MobileCoreInitializer.swift
```diff
@@ -0,0 +1,108 @@
...
+                    MobileCore.configureWith(appId: appId)
+                } else if let filePath = options.filePath {
+                    MobileCore.configureWith(filePath: filePath)
+                }
+
+                // If lifecycleAutomaticTracking flag is false, set lifecycle notification listeners
+                if options.lifecycleAutomaticTrackingEnabled == true {
+                    var usingSceneDelegate = false
```
praveek, line 64: Can you also move this inside setupLifecycle?

AEPCore/Sources/core/MobileCoreInitializer.swift
```diff
@@ -0,0 +1,108 @@
...
+                completion?()
+            }
+        }
+    }
+
+    @available(iOSApplicationExtension, unavailable)
+    @available(tvOSApplicationExtension, unavailable)
+    private func setupLifecycle(usingSceneDelegate: Bool, additionalContextData: [String: Any]? = nil) {
+        if usingSceneDelegate {
+            MobileCore.lifecycleStart(additionalContextData: additionalContextData)
```
praveek, line 86: Is this correct? Shouldn't we trigger lifecycleStart only when the app is foreground?

AEPServices/Sources/utility/ClassFinder.swift
```diff
@@ -0,0 +1,40 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software distributed under
+// the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+// OF ANY KIND, either express or implied. See the License for the specific language
+// governing permissions and limitations under the License.
+//
+
+import Foundation
+
+public struct ClassFinder {
```
cdhoffmann, line 15: Does this need to be Public?

cdhoffmann (approved):

AEPServices/Sources/utility/ClassFinder.swift
praveek: I prefer moving it to Core and making it internal.

AEPCore/Sources/core/MobileCoreInitializer.swift
```diff
@@ -0,0 +1,111 @@
...
+        } else if let filePath = options.filePath {
+            MobileCore.configureWith(filePath: filePath)
+        }
+
+        // Register Extensions
+        DispatchQueue.global().async {
+            let classList = self.classFinder(Extension.self)
+            let filteredClassList = classList.filter { $0 !== AEPCore.EventHubPlaceholderExtension.self && $0 !== AEPCore.Configuration.self }.compactMap { $0 as? NSObject.Type }
+            MobileCore.registerExtensions(filteredClassList) {
+                // If lifecycleAutomaticTracking flag is false, set lifecycle notification listeners
```
praveek, line 61: I also prefer moving the automatic lifecycle tracking before calling registerExtensions so that we don't delay the call. We initially had it this way because in ACP events were dropped before the event hub was initialized, which is not the case with AEP.

AEPCore/Sources/core/MobileCoreInitializer.swift
```diff
@@ -0,0 +1,110 @@
...
+
+        if let appId = options.appId {
+            MobileCore.configureWith(appId: appId)
+        } else if let filePath = options.filePath {
+            MobileCore.configureWith(filePath: filePath)
+        }
+
+        // Register Extensions
+        DispatchQueue.global().async {
+            if options.lifecycleAutomaticTrackingEnabled == true {
```
praveek, line 58: Can you retain [weak self]?

AEPCore/Sources/core/MobileCoreInitializer.swift
```diff
@@ -0,0 +1,110 @@
...
+                self.setupLifecycle(additionalContextData: options.lifecycleAdditionalContextData)
+            } else {
+                Log.trace(label: self.LOG_TAG, "initialize - automatic lifecycle tracking disabled.")
+            }
+
+            let classList = self.classFinder(Extension.self)
+            let filteredClassList = classList.filter { $0 !== AEPCore.EventHubPlaceholderExtension.self && $0 !== AEPCore.Configuration.self }.compactMap { $0 as? NSObject.Type }
```
praveek, line 65: Can you add a comment on why we are filtering these?

AEPCore/Sources/core/MobileCoreInitializer.swift
```diff
@@ -0,0 +1,110 @@
...
+            MobileCore.setAppGroup(options.appGroup)
+        }
+
+        if let appId = options.appId {
+            MobileCore.configureWith(appId: appId)
+        } else if let filePath = options.filePath {
+            MobileCore.configureWith(filePath: filePath)
+        }
+
+        // Register Extensions
```
praveek, line 56: Can you fix the incorrect comment?

AEPCore/Tests/ClassFinderTests.swift
```diff
@@ -0,0 +1,123 @@
...
+@testable import AEPServices
+@testable import AEPCore
+import XCTest
+
+class ClassFinderTests: XCTestCase {
+
+
+    func testClassFinderSimple() {
+        let foundExtensions = ClassFinder.classes(conformToProtocol: Extension.self)
+        XCTAssertEqual(foundExtensions.count, 13) // All Extensions in AEPCore
```
praveek, line 22: Is this retrieving all extensions from the test target? If so, will it fail if I add a new extension for a specific test? I prefer defining a custom protocol and running tests against it to prevent them from breaking.

AEPCore/Tests/MobileCoreInitializerTests.swift
```diff
@@ -0,0 +1,416 @@
...
+
+@testable import AEPCore
+@testable import AEPCoreMocks
+@testable import AEPServices
+
+class MobileCoreInitializerTests: XCTestCase {
+    override func setUp() {
+        NamedCollectionDataStore.clear()
+        MobileCore.resetSDK()
+        MockExtension.reset()
```
praveek, line 24: Can we create a mocked event hub and test the API calls directly, rather than registering a mock extension and adding listeners to it? In general, it would be simpler if we could minimize the number of `EventHub.shared` calls in individual tests?

AEPCore/Sources/core/MobileCore.swift
```diff
@@ -38,6 +39,31 @@ public final class MobileCore: NSObject {
...
+    /// This automatically registers all boundled extensions and sets up lifecycle tracking.
+    /// You can disable automatic lifecycle tracking using `InitOptions`.
+    /// - Parameters:
+    ///   - options: The `InitOptions` used to configure the SDK.
+    ///   - completion: An optional closure triggered once initialization is complete.
+    @available(iOSApplicationExtension, unavailable)
+    @available(tvOSApplicationExtension, unavailable)
+    @objc(initializeWithOptions:completion:)
+    public static func initialize(options: InitOptions, _ completion: (() -> Void)? = nil) {
+        MobileCoreInitializer.shared.initialize(options: options, completion)
```
praveek, line 52: Nit: Given the circular dependency between these two classes, I prefer something like a MobileCore+Initialize extension containing all this logic along with the API, or at least having MobileCore hold a reference to `MobileCoreInitializer` instead of accessing it via `MobileCoreInitializer.shared`. This approach may not scale well when supporting multiple instances.

AEPCore/Tests/ClassFinderTests.swift
kevinlind: I think using a custom Protocol here will be fine. I'll make that change.

praveek (approved):

PR #1121, opened 2025-03-11
Implement event history insert and insertIfNotExists consequence support

timkimadobe:
## Description

This PR implements support for the new `insert` and `insertIfNotExists` consequence support in the Core RulesEngine, leveraging the existing rules condition evaluation logic to trigger these consequences.

The consequence handling has been updated to handle the new `schema` consequence type, and `https://ns.adobe.com/personalization/eventHistoryOperation` schema value.

At a high level, the consequence object has three high level properties:
| Name             | Key       | Type   | Description  |
|-----------------|-----------|--------|-------------|
| Operation type  | operation | string | **Required**<br>Determines the type of database operation to be performed on Event History by the SDK.<br><br> - **Insert**: Inserts a record in the Event History Database.<br> - **Insert if not exists**: Inserts a record in the Event History Database only if there is not already a matching record with a matching hash value. |
| Key-value pairs | content   | object | **Optional**<br>Provided key-value pairs will be added to the string used to generate the event history hash.<br><br>If the value represents a token that the SDK knows how to resolve, the resolved value will be used (such as shared state values, ~timestampu, etc.). |
| Event Data Mask | keys      | array  | **Optional**<br>An array of strings that designate which keys in the event will be used for generating the event history hash. |

The primary logic for this new support is handled in the new `LaunchRulesEngine` `processEventHistoryOperation` method:
1. Checks that the required property of `operation` exists (which could be either `insert` or `insertIfNotExists`, but does not strictly require these two values for future expansion)
2. The consequence's `keys` are separated into two groups - event data keys and token keys: -
    * Token keys: keys that are prefixed with `~` (ex: `~sdkver`)
    * Event data keys: All other keys
3. Token keys are reformatted and resolved
    1. Each token key is placed into a new dictionary where the key is the token and the value is the token wrapped in the LaunchRulesEngine token delimiters
        * `["~sdkver": "{%~sdkver%}"]`
    2 This dictionary is resolved using the `TokenFinder` used to resolve the original rules condition
5. The historical event to be recorded's event data is set up with the original triggering event's data payload
6. The token key dictionary is merged into the data dictionary
7. The consequence's custom `content` key value pairs are merged in (the `content` key value pairs' tokens were already replaced by the overall consequence resolution)
8. The event mask is set up with all of the keys from `keys` (that is, event data keys and token keys) and `content`
9. Create the historical event to be recorded with the merged event data and final mask
10. If `insertIfNotExists`, first check that the hash for the event does not already exist
    * NOTE: A new API, `eventExists` has been created in in `EventHistory` to support easily checking for the hash value instead of having to extract all the key value pairs from the flattened map as is required by the existing `getEvents` API's `EventHistoryRequest` input format requirement
11. Insert the historical event into the record
12. On successful insert, also dispatch the event recorded as a rules consequence event so extensions dependent on event history inserts can respond

## Questions for reviewers
1. `processEventHistoryOperation` uses direct string keys - how do we feel about this? Should we use constants here instead? If so where should they live?:
```swift
if let keys = schemaData["keys"] as? [String] {
```

2. The dispatch consequence event for insert and insertIfNotExists happens inside the processing method itself, not inside the `evaluateRules` loop, because the EventHistory methods it depends on are async methods. Is the approach taken acceptable? Should we also be keeping track of `dispatchChainCount` here?

## Types of changes

## Checklist:

### DIFF
--- a/AEPCore/Sources/rules/LaunchRulesEngine.swift
+++ b/AEPCore/Sources/rules/LaunchRulesEngine.swift
@@ -373,14 +373,16 @@ public class LaunchRulesEngine {
             return
         }
 
-        // Note that `content` doesn't need to be resolved here because it was already resolved by evaluateRules
+        let operationDescription = "'\(operation)' operation for consequence \(consequence.id)"
+
+        // Note `content` doesn't need to be resolved here because it was already resolved by evaluateRules
         guard let content: [String: Any] = schemaData[Self.EVENT_HISTORY_CONTENT_KEY] as? [String: Any] else {
-            Log.warning(label: LOG_TAG, "(\(self.name)) : Unable to process '\(operation)' operation for consequence \(consequence.id) — `content` is either missing or improperly formatted.")
+            Log.warning(label: LOG_TAG, "(\(self.name)) : Unable to process \(operationDescription) — `content` is either missing or improperly formatted.")
             return
         }
 
         if content.isEmpty {
-            Log.warning(label: LOG_TAG, "(\(self.name)) : Unable to process '\(operation)' operation for consequence \(consequence.id) — the consequence `content` does not contain any keys to record.")
+            Log.warning(label: LOG_TAG, "(\(self.name)) : Unable to process \(operationDescription) — the consequence `content` does not contain any keys to record.")
             return
         }
 
@@ -394,35 +396,33 @@ public class LaunchRulesEngine {
         case LaunchRulesEngine.CONSEQUENCE_EVENT_HISTORY_OPERATION_INSERT,
              LaunchRulesEngine.CONSEQUENCE_EVENT_HISTORY_OPERATION_INSERT_IF_NOT_EXISTS:
 
-            // For INSERT_IF_NOT_EXISTS, check if the event exists first
             if operation == LaunchRulesEngine.CONSEQUENCE_EVENT_HISTORY_OPERATION_INSERT_IF_NOT_EXISTS {
                 let eventHash = eventToRecord.eventHash
                 if eventHash == 0 {
-                    Log.warning(label: LOG_TAG, "(\(self.name)) : Unable to process '\(operation)' operation for consequence \(consequence.id) — event hash is 0.")
+                    Log.warning(label: LOG_TAG, "(\(self.name)) : Unable to process \(operationDescription) — event hash is 0.")
                     return
                 }
 
-                // Checking if the event exists before inserting
                 extensionRuntime.getHistoricalEvents([eventToRecord.toEventHistoryRequest()], enforceOrder: false) { [weak self] results in
                     guard let self = self else { return }
                     guard let result = results.first else {
-                        Log.error(label: LOG_TAG, "(\(self.name)) : Unable to get valid historical result for event — skipping '\(operation)' operation for consequence \(consequence.id).")
+                        Log.error(label: LOG_TAG, "(\(self.name)) : Unable to get valid historical result for event — skipping \(operationDescription).")
                         return
                     }
                     if result.count >= 1 {
-                        Log.debug(label: self.LOG_TAG, "(\(self.name)) : Event already exists in history — skipping '\(operation)' operation for consequence \(consequence.id).")
+                        Log.debug(label: self.LOG_TAG, "(\(self.name)) : Event already exists in history — skipping \(operationDescription).")
                         return
                     }
                 }
             }
 
-            Log.trace(label: self.LOG_TAG, "(\(self.name)) : Recording event in history for '\(operation)' operation for consequence \(consequence.id): \(eventToRecord)")
+            Log.trace(label: self.LOG_TAG, "(\(self.name)) : Recording event in history for \(operationDescription): \(eventToRecord)")
             self.extensionRuntime.recordHistoricalEvent(eventToRecord) { [weak self] success in
                 guard let self = self else { return }
                 if success {
                     self.extensionRuntime.dispatch(event: eventToRecord)
                 } else {
-                    Log.warning(label: self.LOG_TAG, "(\(self.name)) : Failed to record event in history for '\(operation)' operation for consequence \(consequence.id).")
+                    Log.warning(label: self.LOG_TAG, "(\(self.name)) : Failed to record event in history for \(operationDescription).")
                 }
             }
 
@@ -430,6 +430,7 @@ public class LaunchRulesEngine {
             Log.warning(label: LOG_TAG, "(\(self.name)) : Unsupported event history operation '\(operation)' for consequence \(consequence.id).")
         }
     }
+
 }
 
 /// Extend RuleConsequence with helper methods for processing Dispatch Consequence events.


AEPCore/Sources/rules/LaunchRulesEngine.swift
```diff
@@ -330,6 +336,123 @@ public class LaunchRulesEngine {
...
     }
+
+    /// Process a schema consequence event. Handles different schema types including event history operations.
+    /// - Parameters:
+    ///   - consequence: the RuleConsequence which contains the schema details
+    ///   - processedEvent: the event that triggered the rule
+    ///   - data: a `Traversable` object used for token resolution
+    private func processSchemaConsequence(consequence: RuleConsequence, processedEvent: Event, data: Traversable) {
+        guard let schema = consequence.schema else {
+            Log.error(label: LOG_TAG, "(\(self.name)) : Unable to process a Schema Consequence, 'schema' is missing from 'details'")
```
spoorthipujariadobe, line 347: nit: As per the [definition](https://wiki.corp.adobe.com/display/ADMSMobile/Schema-based+consequence+details#Schemabasedconsequencedetails-ConsequenceType) of `schema` consequence type, the id and data are also mandatory fields. Wondering if we should have that check here and fast fail

AEPCore/Sources/eventhub/history/EventHistory.swift
```diff
@@ -39,13 +40,35 @@ class EventHistory {
...
+    /// Checks if an `Event` with the same hash exists in the Event History database.
+    ///
+    /// The hash is calculated based on the provided `event`'s data.
+    /// The `event`'s `mask` value, if provided, will filter what values in the event data are used for hash generation.
+    /// If the hash value for the provided `event` is `0`, the method will return `false`.
+    ///
+    /// - Parameters:
+    ///   - event: the `Event` to check for existence in the Event History database
+    ///   - handler: called with a `Bool` indicating if an event with the same hash exists
+    func eventExists(_ event: Event, handler: @escaping (Bool) -> Void) {
```
sbenedicadb, line 60: do we need this function?  don't we achieve the same thing by calling getEvents?

AEPCore/Sources/eventhub/history/EventHistoryDatabase.swift
```diff
@@ -61,6 +61,7 @@ class EventHistoryDatabase {
         dispatchQueue.async {
```
sbenedicadb, line 61: not your code, but can we add a capture of `[weak self]` for any of these async blocks where we are using `self` so we can prevent potential leaking?

AEPCore/Sources/rules/LaunchRulesEngine.swift
```diff
@@ -330,10 +336,130 @@ public class LaunchRulesEngine {
...
     }
+
+    /// Process a schema consequence event. Handles different schema types including event history operations.
+    /// - Parameters:
+    ///   - consequence: the RuleConsequence which contains the schema details
+    ///   - processedEvent: the event that triggered the rule
+    ///   - data: a `Traversable` object used for token resolution
+    private func processSchemaConsequence(consequence: RuleConsequence, processedEvent: Event, data: Traversable) {
+        guard let schema = consequence.schema, let _ = consequence.data, let _ = consequence.detailID else {
+            Log.error(label: LOG_TAG, "(\(self.name)) : Unable to process Schema Consequence with ID: \(consequence.id). A Schema Consequence requires 'schema', 'data', and 'id' properties.")
```
sbenedicadb, line 347: let's change this to a warn please

AEPCore/Sources/rules/LaunchRulesEngine.swift
```diff
@@ -330,10 +336,130 @@ public class LaunchRulesEngine {
...
+            Log.warning(label: LOG_TAG, "(\(self.name)) : Unsupported schema type: \(schema)")
+        }
+    }
+
+    /// Process an event history operation consequence. Records events in the Event History database.
+    /// - Parameters:
+    ///   - consequence: the RuleConsequence which contains the event history operation details
+    ///   - processedEvent: the event that triggered the rule
+    ///   - data: a `Traversable` object used for token resolution
+    private func processEventHistoryOperation(consequence: RuleConsequence, processedEvent: Event, data: Traversable) {
```
sbenedicadb, line 364: let's get the magic strings out of this method: 
- `operation`
- `keys`
- `~`
- `content`

AEPCore/Sources/rules/LaunchRulesEngine.swift
```diff
@@ -330,10 +336,130 @@ public class LaunchRulesEngine {
...
+        guard let schemaData = consequence.data, let operation = schemaData["operation"] as? String else {
+            Log.error(label: LOG_TAG, "(\(self.name)) : Unable to process an EventHistoryOperation Consequence with ID: \(consequence.id). Required 'operation' field missing from 'detail.data'.")
```
sbenedicadb, line 366: please change to `warn`

AEPCore/Sources/rules/LaunchRulesEngine.swift
```diff
@@ -330,10 +336,134 @@ public class LaunchRulesEngine {
...
+                }
+            }
+
+            Log.trace(label: self.LOG_TAG, "(\(self.name)) : Recording event in history with operation '\(operation)'")
+            self.extensionRuntime.recordHistoricalEvent(eventToRecord) { [weak self] success in
+                guard let self = self else { return }
+                if success {
+                    self.extensionRuntime.dispatch(event: eventToRecord)
+                } else {
+                    Log.error(label: self.LOG_TAG, "(\(self.name)) : Failed to record event in history with operation '\(operation)'")
```
sbenedicadb, line 451: warn

sbenedicadb (commented):
looking good so far, a few things to address

AEPCore/Sources/rules/LaunchRulesEngine.swift
```diff
@@ -330,10 +342,138 @@ public class LaunchRulesEngine {
...
+                    eventDataKeys.append(key)
+                }
+            }
+        }
+
+        // Wrap each token key value in the Rules Engine delimiter so they are resolved correctly as tokens
+        var tokenDictionary: [String: Any?] = [:]
+        if !tokenKeys.isEmpty {
+            for key in tokenKeys {
+                tokenDictionary[key] = "\(LaunchRulesEngine.LAUNCH_RULE_TOKEN_LEFT_DELIMITER)\(key)\(LaunchRulesEngine.LAUNCH_RULE_TOKEN_RIGHT_DELIMITER)"
```
spoorthipujariadobe, line 394: On [L173](https://github.com/adobe/aepsdk-core-ios/pull/1121/files#diff-f18c9c144da28be27544a5dd00700ce7e3c76622a25d9c9303020973fabc3887R173), we're traversing through the entire rule consequence and already parsing all tokens inside `{%...%}` and replacing them with values from event data.
Wondering if we should have the place where the rule is authored add in whatever is inside the `keys` part with those delimiters so we don't do this manual addition here and this looks consistent with the other attach and modify consequence . The rule we receive from the server would look like this
```
{
    "id": "123",
    "type": "schema",
    "detail": {
        "id": "abc",
        "schema": "https://ns.adobe.com/personalization/eventHistoryOperation",
        "data": {                              
            "operation": "insert",
            "keys": [
                "{%keyInEventData%}"
            ],
            "content": {
                "key": "value"
            }
        }
    }
}
```
Thoughts?

spoorthipujariadobe: 1. I believe the [LaunchTokenFinder](https://github.com/adobe/aepsdk-core-android/blob/main/code/core/src/main/java/com/adobe/marketing/mobile/launch/rulesengine/LaunchTokenFinder.kt#L29-L42) already has the logic for this today. If the value starts with `~` followed by a known type, like type, source, state, the appropriate value is read. If `~` is followed by an unknown string, we attempt to read from the event data.

2. That would need to be added into `keys` making it a JSON object instead of string of arrays. Example:
```
{
    "id": "123",
    "type": "schema",
    "detail": {
        "id": "abc",
        "schema": "https://ns.adobe.com/personalization/eventHistoryOperation",
        "data": {                              
            "operation": "insert",
            "keys":  {
                "eventType":  "{%~type%}",
                "keyInEventData":  "{%~myEventKey%}
            },
            "content": {
                "key": "value"
            }
        }
    }
}
```

spoorthipujariadobe: Adding the tokens to content would also avoid us calling `EventDataMerger.merge` twice

AEPCore/Sources/eventhub/history/Event+EventHistoryRequest.swift
```diff
@@ -0,0 +1,38 @@
...
+
+        // Filter the flattened data based on mask if provided
+        let filteredData: [String: Any]
+        if let mask = self.mask {
+            // Convert mask array to a set for O(1) lookups
+            let maskSet = Set(mask)
+            filteredData = flattenedData.filter { maskSet.contains($0.key) }
+        } else {
+            // If no mask is provided, use all the data
+            filteredData = flattenedData
```
spoorthipujariadobe, line 33: As per the spec, " One of keys or content must be present in the data object in order for it to be valid.  If the object is invalid, no operation should occur." Should this return null instead of adding all keys?

AEPCore/Sources/eventhub/EventHub.swift
```diff
@@ -356,6 +356,33 @@ final class EventHub {
...
+    /// The event will be recorded based on its calculated hash.
+    /// The hash is generated based on the provided `event`'s data.
+    /// The `event`'s `mask` value, if provided, will filter what values in the event data are used for hash generation.
+    /// If the hash value for the provided `event` is `0`, no record will be created in the database.
+    ///
+    /// - Parameters:
+    ///   - event: the `Event` to be recorded in the Event History database
+    ///   - handler: called with a `Bool` indicating a successful database insert
+    func recordHistoricalEvent(_ event: Event, handler: ((Bool) -> Void)? = nil) {
+        eventHistory?.recordEvent(event, handler: handler)
```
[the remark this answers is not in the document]
sbenedicadb: yup, seems good

AEPCore/Sources/eventhub/history/EventHistory.swift
sbenedicadb: perfect - i think that's a great plan 👍

AEPCore/Sources/eventhub/history/Event+EventHistoryRequest.swift
spoorthipujariadobe: We already have the logic to check if `content` is [non empty](https://github.com/adobe/aepsdk-core-ios/pull/1121/files#diff-f18c9c144da28be27544a5dd00700ce7e3c76622a25d9c9303020973fabc3887R384-R387). But it might be good to also add that here since we cannot always be sure that the caller will have the check. As per the spec, only the keys which exist in the `mask` array are recorded in event history. If the `mask` is empty, nothing is recorded in event history. Is my understanding correct?

spoorthipujariadobe: Discussed offline, using all keys to generate hash in case of no mask aligns with existing behavior

spoorthipujariadobe (approved):
LGTM! Lets add tests to cover the new rule consequences

PR #1124, opened 2025-04-15
-add in support for maxWidth to MessageSettings and use it when calculating IAM frames

sbenedicadb:
## Types of changes

## Checklist:


sbenedicadb:
@spoorthipujariadobe pointed out we needed a minor version update for this release, so i've update the base branch to be 5.5.0

AEPServices/Sources/ui/fullscreen/FullscreenMessage+FrameCalculation.swift
```diff
@@ -119,13 +119,17 @@
         /// width in settings represents a percentage of the screen.
         /// e.g. - 80 = 80% of the screen width.
```
cdhoffmann, line 120: Are we changing the definition of width now to not be a percentage if maxWidth is set? If so, we should update the definition here as its a bit confusing.

AEPServices/Sources/ui/fullscreen/FullscreenMessage.swift
```diff
@@ -495,6 +495,19 @@
                 }
             }
         }
+
+        // MARK: - debug methods below are used for testing purposes only
```
cdhoffmann, line 499: Can we move all the debug functions to their own debug extension in a separate file?

cdhoffmann (changes requested):

AEPServices/Sources/ui/fullscreen/FullscreenMessage+FrameCalculation.swift
sbenedicadb: no. this comment is referring to the value in settings.  i'll add a comment that explains the broader purpose of the calculated variable for clarity (and for height as well)

AEPServices/Sources/ui/fullscreen/FullscreenMessage.swift
sbenedicadb: i don't think we can - if we move them to a separate file then they won't be able to expose the underlying private methods (which is what we're trying to achieve to begin with).

PR #1128, opened 2025-04-21
[MOB-23277] Deprecate getCarrierName on iOS 16

cdhoffmann:
## Types of changes

## Checklist:


AEPServices/Sources/SystemInfoService.swift
```diff
@@ -37,6 +37,7 @@ public protocol SystemInfoService {
 
     /// Gets the mobile carrier name
     /// - Return: `String` the mobile carrier name
+    @available(iOS, deprecated: 16.0, message: "Apple has deprecated the CTCarrier API without replacement.")
```
kevinlind, line 40: Will this now cause a deprecation warning for applications using Lifecycle running on iOS 16?

cdhoffmann: Yes. Lmk if you think i should remove the warning or not. I added it because this API is useless on iOS 16 and above.

kevinlind: I think it best to remove the warning as we are not planning on removing the API from the SDK. Though, I'd suggest updating the comment to state the API always returns nil on iOS 16 and above because Apple has deprecated the CTCarrier API without replacement.

PR #1132, opened 2025-04-21
Add tests for new insert rules consequence (part 1)

timkimadobe:
## Description

This PR adds test cases for: 

### EventHub's EventHistory related methods
```swift
func getHistoricalEvents(_: [EventHistoryRequest], enforceOrder: Bool, handler: @escaping ([EventHistoryResult]) -> Void)
func recordHistoricalEvent(_: Event, handler: ((Bool) -> Void)? = nil)
```

One big motivation for adding test coverage here was the update to change EventHub's implementation of `getHistoricalEvents` in PR:
* https://github.com/adobe/aepsdk-core-ios/pull/1121

Where there were no test cases validating the behavior of the handler being called or not in the `eventHistory` is `nil` case.

**Before**
```swift
eventHistory?.getEvents(requests, enforceOrder: enforceOrder, handler: handler)
```

**After**
```swift
eventHistory?.getEvents(requests, enforceOrder: enforceOrder, handler: handler) ?? handler([])
```

### `Event`'s new support for conversion to `EventHistoryRequest` via `.toEventHistoryRequest`
Tests the conversion of `Event` into an `EventHistoryRequest` especially when different event masks are set.

### Other changes
New testing initializers for `EventHistory` and `EventHistoryDatabase`, so that the classes can be initialized with mocks and not have to call production init logic. Also added clear documentation that changes to init logic need to be reflected in the test initializers.
* The initializers are also wrapped in `#if DEBUG` so they cannot be mistakenly used in production
* Although this is an inversion of the dependency, I thought the alternative of creating a new protocol to separate the concrete implementations from the methods was too big of a lift to accommodate this case.
* Considered alternative of using delegated initialization via `convenience init`, but **`convenience init` is not accessible to classes inheriting from the base class.** Due to EventHub using the classes directly, in order to conform to the type requirement, mock classes need to inherit from the same class. This means having to define two initializers anyway; a private shared delegated init and an internal DEBUG init:
```swift
#if DEBUG
init(db: EventHistoryDatabase) {
    self.db = db
}
#else
private init(db: EventHistoryDatabase) { // <-- basically DEBUG is just to control this access level
    self.db = db
}
#endif

convenience init?() {
    guard let db = EventHistoryDatabase(dispatchQueue: DispatchQueue.init(label: "EventHistory")) else {
        return nil
    }
    self.init(db: db)
}
```

## Types of changes

## Checklist:

### DIFF
--- a/AEPCore.xcodeproj/project.pbxproj
+++ b/AEPCore.xcodeproj/project.pbxproj
@@ -334,8 +334,8 @@
 		3FF829492507F8AA00483C74 /* rules_lifecycle.json in Resources */ = {isa = PBXBuildFile; fileRef = 3FF829482507F8AA00483C74 /* rules_lifecycle.json */; };
 		3FF829692509937100483C74 /* SignalIntegrationTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3FF829682509937100483C74 /* SignalIntegrationTests.swift */; };
 		3FF8296D2509942300483C74 /* rules_signal.json in Resources */ = {isa = PBXBuildFile; fileRef = 3FF8296B2509942300483C74 /* rules_signal.json */; };
-		4C0111D12DB1C3F000996757 /* EventHistorySpy.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4C0111D02DB1C3F000996757 /* EventHistorySpy.swift */; };
-		4C0111D32DB1C4FD00996757 /* EventHubHistoryTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4C0111D22DB1C4FD00996757 /* EventHubHistoryTests.swift */; };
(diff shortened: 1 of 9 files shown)


AEPCore/Tests/EventHubTests/HistoryTests/Event+EventHistoryRequestTests.swift
```diff
@@ -0,0 +1,80 @@
...
+ OF ANY KIND, either express or implied. See the License for the specific language
+ governing permissions and limitations under the License.
+*/
+
+import XCTest
+
+@testable import AEPCore
+
+class Event_EventHistoryRequestTests: XCTestCase {
+    func testToEventHistoryRequest_dataNilAndMaskNil_returnsEmptyMask() {
```
spoorthipujariadobe, line 18: nit: is the test name supposed to say `returnEmptyEventHistoryRequest`?

spoorthipujariadobe (approved):

PR #1134, opened 2025-04-23
Fix event history operation `insertIfNotExists` logical error and race with insert operation

timkimadobe:
## Description

This PR fixes `insertIfNotExists` handling logic by addressing the following issues:
1. A logical error where the early exit cases for `insertIfNotExists` return from the escaping closure, but those returns do not actually affect the main `processEventHistoryOperation` block that would continue to do the event insert into event history.
2. A race between the `insertIfNotExists` event history record check which uses an escaping closure in `getHistoricalEvents` and the insertion of the event into event history using `recordHistoricalEvent`
3. Also added handling for the `insertIfNotExists` case where a database error value (-1) is returned, skipping the insert if an error occurs.  
   * I was thinking that the logic behind checking for records is to affirmatively insert, so if the check fails, it would be safer to default to not inserting, rather than inserting regardless, as inserting in this case could violate the utility provided by the `IfNotExists` portion of the logic.
4. For the logical checks and related logging in `insertIfNotExists`, I chose to maintain more granularity in the log messages rather than condensing all the checks into a single guard, because I felt that providing non-technical explanations (ex: just the result code) for why an insert is not happening makes it much easier to self-service troubleshooting for workflows that rely on this feature. 
    * Basically, in weighing the tradeoff between the maintenance cost of the code and the ease of understanding the debugging workflow for those outside of our team, I leaned more towards prioritizing the latter in this case

## Alternative
Could more simply add the insert separately for both operations instead of having them merge back into the same insert path.
* `insert` would have its own conditional block checking separately for this specific operation instead of sharing logic with `insertIfNotExists`
* `insertIfNotExists` would have "insert event into event history" logic separately inside its escaping closure

However, this would change the behavior of rules consequences when `insertIfNotExists` is used such that when `insertIfNotExists` consequences are used, their effects are no longer guaranteed to: 
1. Complete running in the same consequence order as defined in a given rule's consequences, nor
2. Complete running within the context of the rule condition it was defined under.

Given these knock on effects, I elected to make the `getHistoricalEvents` synchronous instead

**Example**
```swift
// Inner function that handles shared insert logic
func insert(event: Event) {
    Log.trace(label: LOG_TAG, "\(logPrefix) - Recording event in history with operation '\(operation)'")

    // For INSERT and INSERT_IF_NOT_EXISTS (which passed not exists check), insert the event
    self.extensionRuntime.recordHistoricalEvent(event) { [weak self] success in
        if success {
            self.extensionRuntime.dispatch(event: event)
        } else {
            Log.warning(label: LOG_TAG, "\(logPrefix) - Failed to record event in history for '\(operation)'")
        }
    }
}
// insertIfNotExists handling
if operation == LaunchRulesEngine.CONSEQUENCE_EVENT_HISTORY_OPERATION_INSERT_IF_NOT_EXISTS {
    let eventHash = eventToRecord.eventHash
    if eventHash == 0 {
        Log.warning(label: LOG_TAG, "\(logPrefix) - Unable to process '\(operation)' operation, event hash is 0")
        return
    }

    extensionRuntime.getHistoricalEvents([eventToRecord.toEventHistoryRequest()], enforceOrder: false) { [weak self] results in

        guard let result = results.first else {
            Log.warning(label: LOG_TAG, "\(logPrefix) - Unable to retrieve historical events, skipping '\(operation)' operation")
            return
        }
        if result.count >= 1 {
            Log.trace(label: LOG_TAG, "\(logPrefix) - Event already exists in history, skipping '\(operation)' operation")
            return
        }
        insert(event: eventToRecord)
    }
}
// insert handling
if operation == LaunchRulesEngine.CONSEQUENCE_EVENT_HISTORY_OPERATION_INSERT {
    insert(event: eventToRecord)
}
```

## Types of changes

## Checklist:

### DIFF
--- a/AEPCore/Sources/rules/LaunchRulesEngine.swift
+++ b/AEPCore/Sources/rules/LaunchRulesEngine.swift
@@ -406,7 +406,9 @@ public class LaunchRulesEngine {
 
                 // Check that a valid result exists and that a database error (-1) was not returned
                 guard let result = results.first, result.count >= 0 else {
-                    Log.warning(label: LOG_TAG, "\(logPrefix) - Unable to retrieve historical events, skipping '\(operation)' operation. Returned with result value: \(String(describing: results.first?.count))")
+                    Log.warning(label: LOG_TAG,
+                        "\(logPrefix) - Unable to retrieve historical events, skipping '\(operation)' operation. " +
+                        "Returned with result value: \(String(describing: results.first?.count))")
                     return
                 }
                 if result.count >= 1 {


AEPCore/Sources/rules/LaunchRulesEngine.swift
```diff
@@ -406,12 +413,23 @@ public class LaunchRulesEngine {
                     Log.trace(label: LOG_TAG, "\(logPrefix) - Event already exists in history, skipping '\(operation)' operation")
                     return
                 }
+                eventExistsInEventHistory = false
+            }
+            let timeoutResult = semaphore.wait(timeout: .now() + .seconds(LaunchRulesEngine.EVENT_HISTORY_LOOKUP_TIMEOUT_SEC))
```
spoorthipujariadobe, line 418: We discussed offline yesterday about potentially removing this time out and letting the lookup run since its happening on a separate rules engine thread. Are we planning to keep it or remove it?

spoorthipujariadobe (approved):
LGTM, just one question around the timeout so I can make the same changes on Android

AEPCore/Sources/rules/LaunchRulesEngine.swift
```diff
@@ -40,6 +40,7 @@ public class LaunchRulesEngine {
     // Event History Operation constants
     private static let EVENT_HISTORY_OPERATION_KEY = "operation"
     private static let EVENT_HISTORY_CONTENT_KEY = "content"
+    private static let EVENT_HISTORY_LOOKUP_TIMEOUT_SEC = 1
```
spoorthipujariadobe, line 43: nit: Let's remove this constant as well since its no longer used

spoorthipujariadobe (approved):

PR #1135, opened 2025-04-24
Add tests for new insert rules consequence (part 2)

timkimadobe:
## Description

This PR adds the tests covering the historical event operations `insert` and `insertIfNotExists`. At a high level, it validates:
* Correct consequence format for all properties in the newly handled schema type
* Event recorded is what was defined in the consequence's `content`
* Event recorded is the same as what is dispatched
* `insertIfNotExists` correctly does not insert if event already exists or a database error is returned (-1)

### Test setup changes
The base test setup for `RulesEngineFunctionalTests` has been split into a new base class: `RulesEngineTestBase`, which the new `RulesEngineHistoricalTests` also uses.

The `TestableExtensionRuntime` has been updated to support:
* Capturing the event to be recorded, when passed through `recordHistoricalEvent`
* Setting an optional delay for the `getHistoricalEvents` operation (by default does not apply any delay)

## Types of changes

## Checklist:

### DIFF
--- a/AEPCore.xcodeproj/project.pbxproj
+++ b/AEPCore.xcodeproj/project.pbxproj
@@ -3039,13 +3039,9 @@
 			inputFileListPaths = (
 				"${PODS_ROOT}/Target Support Files/Pods-AEPLifecycleTests/Pods-AEPLifecycleTests-frameworks-${CONFIGURATION}-input-files.xcfilelist",
 			);
-			inputPaths = (
-			);
 			name = "[CP] Embed Pods Frameworks";
 			outputFileListPaths = (
 				"${PODS_ROOT}/Target Support Files/Pods-AEPLifecycleTests/Pods-AEPLifecycleTests-frameworks-${CONFIGURATION}-output-files.xcfilelist",
-			);
-			outputPaths = (
 			);
 			runOnlyForDeploymentPostprocessing = 0;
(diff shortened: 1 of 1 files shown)


AEPCore/Tests/FunctionalTests/RulesEngineHistoricalTests.swift
```diff
@@ -0,0 +1,679 @@
...
+        // When:
+        rulesEngine.process(event: defaultEvent)
+
+        // Then: Event history consequence event:
+        // Should not attempt to record into event history
+        XCTAssertFalse(mockRuntime.recordHistoricalEventCalled)
+        // Should not be dispatched
+        XCTAssertEqual(0, mockRuntime.dispatchedEvents.count)
+    }
+}
```
spoorthipujariadobe, line 679: nit : Can we also add a test case for `testSchemaConsequenceInsert_doesNotDispatchOrRecord_whenContentIsEmpty`

spoorthipujariadobe (approved):

AEPCore/Tests/FunctionalTests/RulesEngineHistoricalTests.swift
spoorthipujariadobe: Ah I missed it. Should be fine where it is now too

PR #1136, opened 2025-04-24
Create protocol layers for EventHistory and EventHistoryDatabase

timkimadobe:
## Description

This PR creates protocol layers for EventHistory and EventHistoryDatabase for a cleaner testing setup.

The new protocols, `EventHistoryService` and `EventHistoryDatabaseService`, simply define the current internally available methods from their respective classes.

The existing `EventHistory` and `EventHistoryDatabase` classes have been updated to conform to their respective protocols.

`EventHub` has been updated to use the `EventHistoryService` protocol for `eventHistory` and use dependency injection to set this value, with the default being the real `EventHistory`
* By providing the real `EventHistory` as the default value, it doesn't change the signature of any previous calls to `EventHub()`
> [!NOTE] 
> My understanding (please correct me if I'm wrong!) is that this also shouldn't change any real, meaningful runtime behavior of `EventHub`, since by moving the instantiation of `EventHistory` from the default property value to the initializer, `EventHistory` creation and set is being moved from phase 1 to phase 2 of class initialization as described here: https://docs.swift.org/swift-book/documentation/the-swift-programming-language/initialization/#Two-Phase-Initialization
>    * Which means that the initialization and setting of the real `EventHistory` still happens completely within the initialization process of `EventHub`

## Types of changes

## Checklist:

### DIFF
--- a/AEPCore.xcodeproj/project.pbxproj
+++ b/AEPCore.xcodeproj/project.pbxproj
@@ -2936,13 +2936,9 @@
 			inputFileListPaths = (
 				"${PODS_ROOT}/Target Support Files/Pods-AEPLifecycleTests/Pods-AEPLifecycleTests-frameworks-${CONFIGURATION}-input-files.xcfilelist",
 			);
-			inputPaths = (
-			);
 			name = "[CP] Embed Pods Frameworks";
 			outputFileListPaths = (
 				"${PODS_ROOT}/Target Support Files/Pods-AEPLifecycleTests/Pods-AEPLifecycleTests-frameworks-${CONFIGURATION}-output-files.xcfilelist",
-			);
-			outputPaths = (
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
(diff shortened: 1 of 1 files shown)


cdhoffmann (approved):
Looks great!

prudrabhat (approved):

AEPCore/Sources/eventhub/EventHub.swift
```diff
@@ -34,7 +34,7 @@ final class EventHub {
     private let eventQueue = OperationOrderer<Event>("EventHub")
     private var preprocessors = ThreadSafeArray<EventPreprocessor>(identifier: "com.adobe.eventHub.preprocessors.queue")
     private var started = false // true if the `EventHub` is started, false otherwise. Should only be accessed from within the `eventHubQueue`
-    private var eventHistory = EventHistory()
+    private var eventHistory: EventHistoryService?
```
prudrabhat, line 37: Can this be made to a Kotlin `final`/`val` equivalent

PR #1138, opened 2025-04-29
Implement `mostRecent` search type for rules historical condition

timkimadobe:
## Description

This PR implements the `mostRecent` search type for rules historical condition, to support the requalification workflow.

The `mostRecent` search type is defined as follows:

>The return type for this search is an integer. It represents the 0-based index of the most recent event, or -1 if none of the events are present in the event history.
>
>The formats of the `events` parameter and its anonymous objects are the same as currently defined for historical conditions.
>
> For example, given three distinct events in the events parameter: A, B, and C:
>
> events: [A, B, C]
>
> The system searches for all three events in the event history and checks which event is the most recent. For example, if B is the most recent event, the return value is 1. If A is the most recent event, the return value is 0.

A new Operand function called `getMostRecentHistoricalEvent` has been created to perform the most recent search, which requires as inputs:
* `runtime: ExtensionRuntime` - to call getHistoricalEvents
* `requestEvents: [EventHistoryRequest]` - to know which events to compare to return most recent

For each event provided in the condition, it searches event history and returns the index of the event with the newest occurrence, or -1 if none exist or a database error was encountered.
* This early exit and return -1 is adopted from the existing `getHistoricalEventCount` implementation

Also adds new warnings for early exit or fallback paths in the related `extractHistoricalCondition` method

### Alternative implementations
Considered potentially concurrently dispatching the `getHistoricalEvents` requests. Doesn't make sense for two reasons:
1. First and foremost all database requests go through EventHistoryDatabase's serial queue, which means even if requests arrive concurrently, they are processed synchronously from the perspective of the database
2. Even if the first point wasn't the case, it is an explosion of complexity for very little gain - currently the most complex use case for this logic is 2 events at most (qualify, unqualify)

## Types of changes

## Checklist:

### DIFF
--- a/AEPCore/Sources/rules/JSONRulesParser.swift
+++ b/AEPCore/Sources/rules/JSONRulesParser.swift
@@ -86,13 +86,6 @@ enum EventHistorySearchType: String, Codable {
     case any
     case mostRecent
     case ordered
-
-    init?(_ optionalValue: String?) {
-        guard let value = optionalValue, let enumValue = EventHistorySearchType(rawValue: value) else {
-            return nil
-        }
-        self = enumValue
-    }
 }
 
 class JSONCondition: Codable {
@@ -168,13 +161,9 @@ class JSONCondition: Codable {
         if let toTs = definition.to {
             toDate = Date(milliseconds: toTs)
         }
-        // Attempt to convert to a valid EventHistorySearchType.
-        // If conversion fails, log a warning and fall back to .any.
-        let searchType = EventHistorySearchType(definition.searchType) ?? {
-            Log.warning(label: JSONRulesParser.LOG_LABEL,
-                        "Unknown historical condition searchType '\(definition.searchType ?? "nil")', falling back to .any")
-            return .any
-        }()
+
+        // Default search type is `.any`. Replaced by a valid type from `definition.searchType` if provided.
+        let searchType = definition.searchType.flatMap(EventHistorySearchType.init) ?? .any
 
         let requestEvents = events.map({
             EventHistoryRequest(mask: $0, from: fromDate, to: toDate)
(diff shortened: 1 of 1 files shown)


AEPCore/Sources/rules/JSONRulesParser.swift
```diff
@@ -189,17 +213,16 @@ class JSONCondition: Codable {
         let semaphore = DispatchSemaphore(value: 0)
 
         runtime.getHistoricalEvents(requestEvents, enforceOrder: searchType == .ordered) { results in
+            defer { semaphore.signal() }
```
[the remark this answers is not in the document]
sbenedicadb: i'm with you on the "don't fix what's not broken" change.  thoughts on both suggestions though:
1. the defer looks like it's in the right spot and shouldn't change any functionality (our tests should confirm this!) 
2. i'm totally fine with a param count check on 205 - it doesn't change anything functionally but could serve as a fast fail if we don't have at least those 3 parameters we need.

similar to the comment made earlier in the review, i'm thinking that by using a switch statement inside of this callback we'd add some clarity to someone reading the code.  what do you think about the idea of a little refactor?

AEPCore/Sources/rules/JSONRulesParser.swift
```diff
@@ -210,21 +233,65 @@ class JSONCondition: Codable {
...
             }
-
-            semaphore.signal()
         }
 
         semaphore.wait()
         return returnValue
     }
 
+    /// Returns the index of the most recent historical event based on occurrence date.
```
sbenedicadb, line 248: nit: `date` > `timestamp`

AEPCore/Sources/rules/JSONRulesParser.swift
```diff
@@ -210,21 +233,65 @@ class JSONCondition: Codable {
...
+        }
+
+        var mostRecentIndex = -1
+        var mostRecentDate = Date.distantPast
+        let semaphore = DispatchSemaphore(value: 0)
+
+        runtime.getHistoricalEvents(requestEvents, enforceOrder: false) { eventResults in
+            defer { semaphore.signal() }
+
+            for (index, result) in eventResults.enumerated() {
```
sbenedicadb, line 272: clever 🤩

sbenedicadb (changes requested):

AEPCore/Sources/rules/JSONRulesParser.swift
```diff
@@ -179,52 +191,104 @@ class JSONCondition: Codable {
...
+    /// If duplicate ``EventHistoryRequest``s are provided, the index of the first instance will be returned.
+    ///
+    /// - Parameter parameters: An array expected to contain, in order:
+    ///     - An ``ExtensionRuntime`` instance used to perform historical event lookups.
+    ///     - An array of ``EventHistoryRequest`` objects specifying the search criteria.
+    ///
+    /// - Returns: The index of the event with the most recent occurrence, or `-1` if none are found or there is an error.
+    func getMostRecentHistoricalEvent(parameters: [Any]?) -> Int {
+        guard let params = parameters,
+              params.count >= 2,
```
spoorthipujariadobe, line 256: iOS novice question: On L172, params passed into `getMostRecentHistoricalEvent` is an array of three elements, same as `getHistoricalEventCount`. Won't this check fail?

PR #1143, opened 2025-05-07
Create Decodable helper method

timkimadobe:
## Description

This PR creates a helper method on Decodable which converts a dictionary into an instance of that Decodable type; basically the opposite of the existing Encodable method `asDictionary()`

## Alternative considered
Creating an `.aep` namespaced Codable helpers which help:
* Convert Encodable instances into `[String: Any]` and
* Convert `[String: Any]` into Decodable instances

The motivation for the namespace was to avoid polluting the global namespace, especially when attached to very high usage protocols like `Encodable` and `Decodable`

The existing Encodable helper method `asDictionary` would have been marked as deprecated and the associated warning also provide a one click fix-it to migrate to the new namespaced helper:

**Before:**

Screenshot 2025-05-07 at 1 42 30 PM

**After:** 

Screenshot 2025-05-07 at 1 42 37 PM

## Types of changes

## Checklist:

### DIFF
--- a/API/AEPServices-ios.json
+++ b/API/AEPServices-ios.json
@@ -19234,6 +19234,89 @@
       },
       {
         "kind": "TypeDecl",
+        "name": "Decodable",
+        "printedName": "Decodable",
+        "children": [
+          {
+            "kind": "Function",
+            "name": "fromDictionary",
+            "printedName": "fromDictionary(_:dateDecodingStrategy:)",
+            "children": [
+              {
+                "kind": "TypeNominal",
+                "name": "Optional",
+                "printedName": "τ_0_0?",
+                "children": [
+                  {
+                    "kind": "TypeNominal",
+                    "name": "GenericTypeParam",
+                    "printedName": "τ_0_0"
+                  }
+                ],
+                "usr": "s:Sq"
+              },
+              {
+                "kind": "TypeNominal",
+                "name": "Dictionary",
+                "printedName": "[Swift.String : Any]",
+                "children": [
+                  {
+                    "kind": "TypeNominal",
+                    "name": "String",
+                    "printedName": "Swift.String",
+                    "usr": "s:SS"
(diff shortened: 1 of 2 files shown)


cdhoffmann (changes requested):
Looks great, just a couple of things. 

Can we update all of our uses of the asDictionary API to use the new namespace? 

Also, i am not 100% sold on the naming. idk why but the all lowercase aep just feels off to me. That being said i am going to let the naming KING @sbenedicadb decide the fate of the name.

AEPServices/Sources/utility/AnyCodable.swift
```diff
@@ -251,11 +251,38 @@ extension AnyCodable: Equatable {
...
+
+extension Decodable {
+    /// Creates an instance of `Self` from a `[String: Any]` dictionary.
+    ///
+    /// - Parameters:
+    ///   - dictionary: The source dictionary.
+    ///   - dateDecodingStrategy: How `Date` values should be decoded.
+    ///     Defaults to `.deferredToDate`.
+    /// - Returns: A fully-decoded instance of `Self`, or `nil` if decoding fails.
+    static func fromDictionary(
```
sbenedicadb, line 275: where all do we intend to use this?  only inside of core?  just asking because i noticed it's not public and wondered if that was intentional or not :)

sbenedicadb (approved):
looks good pending the possible update for access control

cdhoffmann (approved):

AEPServices/Sources/utility/AnyCodable.swift
PravinPK: @timkimadobe  I think this was added to be used in Live Activity code (Messaging extension)
Do you modify the access to public ?

PravinPK:
Changed the base branch to v5.5.1

PR #1154, opened 2025-06-23
[PLATIR-51473] Fix FullScreenMessage onDismiss being called before dismissal

cdhoffmann:
We were calling onDismiss for the `FullScreenMessageDelegate` and the MessagingDelegate within the `dismiss()` function before we actually dismissed the message within the  `dismissWithAnimation`. This meant that if someone was listening for the `onDismiss` call while managing their `FullScreenMessage`, they could delete the instance of the FullScreenMessage in between the `dismiss` and `dismissWithAnimation` call. In addition, the `dismissWithAnimation` function recently added a [weak self] and guard self = self else { return } which would mean that self could be nil here and the actual dismiss code would never be reached. 

## Description

Now, the onDismiss delegate calls will happen after the dismissal actually occurs and only when the `dismissWithAnimation(shouldDeallocateWebView: Bool)` parameter is true because we also call this within `hide()`, and we do not want to call onDismiss when hiding. 

## Related Issue

[PLATIR-51473]

## Types of changes

## Checklist:

- [x] I have signed the [Adobe Open Source CLA](http://opensource.adobe.com/cla.html).
- [x] My code follows the code style of this project.

### DIFF
--- a/AEPServices/Sources/ui/fullscreen/FullscreenMessage.swift
+++ b/AEPServices/Sources/ui/fullscreen/FullscreenMessage.swift
@@ -256,10 +256,6 @@
 
                 self.dismissWithAnimation(shouldDeallocateWebView: true)
 
-                // notify all listeners
-                self.listener?.onDismiss(message: self)
-                self.messagingDelegate?.onDismiss(message: self)
-
                 // remove the temporary html if it exists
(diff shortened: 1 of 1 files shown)


sbenedicadb (approved):
looks good.  thanks.

Issue #1043, opened 2024-06-04
Fix a bug in LaunchRulesEngine which causes some events to be processed twice 


praveek:
### Task description

Fix a race condition where rules are processed twice for events dispatched between replaceRules() and before the rules engine processing `rulesEngine:reset` event.

PR #1176, opened 2025-08-15
Fix race condition in rules engine which caused some events to be processed twice

praveek:
Issue: https://github.com/adobe/aepsdk-core-ios/issues/1043
Fixes an issue where cached events occurring between replaceRules and the resetRules event being received by the rules engine are processed twice.

Improve Identity extension by eliminating the shared state checks for events unrelated to the Identity extension.


kevinlind (approved):

praveek:
Closing this for now, will reopen after extension tests.

PR #1195, opened 2025-12-23
Added reEvaluation support for rules engine

akhiljain1907:
## Description
This PR adds support for parsing the meta object with reEvaluate key in rules JSON and introduces a re-evaluation mechanism for the Launch Rules Engine.

Changes
1. Meta Object Parsing
- LaunchRule now includes a reevaluable property parsed from meta.reEvaluate in the rules JSON
- Added JSONRuleMeta struct to handle metadata parsing
- Defaults to false if not specified

2. Re-evaluation Interceptor Support
- Extensions using LaunchRulesEngine can register a RuleReevaluationInterceptor via setReevaluationInterceptor(_:)
- When reevaluable rules match, the interceptor is notified and rule processing is paused
- Processing resumes when the extension calls the completion callback

3. Schema Consequence Handling
- Only rules with schema type consequences can trigger re-evaluation
- When any reevaluable rule matches, all rules with schema consequences are held back
- This ensures consistency in consequence delivery and behavior
- Non-schema rules (add, mod, dispatch, etc.) are processed immediately

## Types of changes

## Checklist:


AEPCore/Sources/rules/LaunchRule.swift
```diff
@@ -17,9 +17,11 @@ import AEPRulesEngine
 public struct LaunchRule: Rule {
     public let condition: Evaluable
     public let consequences: [RuleConsequence]
+    public let reevaluable: Bool
```
sbenedicadb: i think it's better to pull in the `meta` dictionary here rather than one specific value inside that dictionary.

this puts us in a better place for future usage of the `meta` object inside of a launch rule to deliver new properties, etc.

we could make `reEvaluable` a read-only computed property for convenience.

AEPCore/Sources/rules/JSONRulesParser.swift
```diff
@@ -63,17 +63,36 @@ struct JSONRuleRoot: Codable {
...
     }
 }
 
+/// Metadata for a rule containing reevaluation settings
+struct JSONRuleMeta: Codable {
+    var reEvaluate: Bool?
+    
+    enum CodingKeys: String, CodingKey {
+        case reEvaluate
+    }
```
sbenedicadb: don't need coding keys enum since the property name we are using is the same as the named coding key

AEPCore/Sources/rules/JSONRulesParser.swift
```diff
@@ -63,17 +63,36 @@ struct JSONRuleRoot: Codable {
...
                 result.append(rule)
             }
         }
         return result
     }
 }
 
+/// Metadata for a rule containing reevaluation settings
+struct JSONRuleMeta: Codable {
+    var reEvaluate: Bool?
```
sbenedicadb: might be more clear to a reader and safer to just make this value non-optional, and with a default value?

```
let reEvaluate: Bool = false

sbenedicadb (changes requested):
looks good with just a few updates requested 👍

AEPCore/Sources/rules/LaunchRule.swift
akhiljain1907: Yes, we should have the meta dictionary passed instead of just reEvaluate key for future usage. I have made that change.

AEPCore/Sources/rules/JSONRulesParser.swift
akhiljain1907: Yes i have removed it.

AEPCore/Sources/rules/JSONRulesParser.swift
akhiljain1907: As i am parsing entire meta dictionary now, i have removed this JSONRuleMeta struct.

akhiljain1907:
Thanks a lot @sbenedicadb for reviewing. I have addressed the comments, can you please take a look and approve if it looks good?

sbenedicadb (approved):
looks good!

ishwetansh (approved):

sachauhan-adobe (approved):

PR #1196, opened 2026-02-09
Dev v5.8.0 to Staging 

akhiljain1907:
## Types of changes

## Checklist:


cdhoffmann (approved):

AEPCore/Sources/rules/LaunchRule+Reevaluable.swift
```diff
@@ -0,0 +1,42 @@
+/*
+ Copyright 2020 Adobe. All rights reserved.
```
sbenedicadb: new file copyrights must include the current year.  please change to 2026

AEPCore/Sources/rules/LaunchRule+Reevaluable.swift
```diff
@@ -0,0 +1,42 @@
...
+extension LaunchRule {
+    
+    /// Consequence types that support re-evaluation.
+    /// Currently, only "schema" consequences are considered re-evaluable.
+    static let reevaluableConsequenceTypes: Set<String> = ["schema"]
+    
+    /// Indicates whether this rule should trigger re-evaluation when matched.
+    /// Reads from `meta["reEvaluate"]`, defaults to `false` if not present.
+    public var reevaluable: Bool {
+        return meta?["reEvaluate"] as? Bool ?? false
```
sbenedicadb: nit: move both of these magic strings to the constants file

AEPCore/Sources/rules/LaunchRulesEngine.swift
```diff
@@ -14,6 +14,22 @@ import AEPServices
...
+    /// - Parameters:
+    ///   - event: The event that triggered the rules
+    ///   - reevaluableRules: Rules marked as reevaluable
+    ///   - completion: Call when done to trigger re-evaluation
+    func onReevaluationTriggered(
+        event: Event,
+        reevaluableRules: [LaunchRule],
+        completion: @escaping () -> Void
+    )
+}
```
sbenedicadb: nit: would prefer having this protocol defined in its own file.

sbenedicadb (changes requested):
a couple very small requests

AEPCore/Sources/rules/LaunchRulesEngine.swift
```diff
@@ -46,6 +62,9 @@ public class LaunchRulesEngine {
     private let rulesQueue: DispatchQueue
     private var waitingEvents: [Event]?
     private var dispatchChainedEventsCount: [UUID: Int] = [:]
+    
+    /// The interceptor that handles reevaluation when reevaluable rules match
+    private weak var reevaluationInterceptor: RuleReevaluationInterceptor?
```
sbenedicadb, line 53: i'm having second thoughts on making this a `weak` reference.  it assumes that the containing app will maintain a reference to this object throughout the duration of the app, and i worry it will lead to unexpected no-ops (which is what would happen if `reevaluationInterceptor` is nil)

wdyt?

sbenedicadb: if we do decide to make this a strong reference, let's also update the protocol, removing the limitation that it has to be a class to implement it (remove `: AnyObject` from the protocol definition)

sbenedicadb (approved):

PR #1225, opened 2026-09-04
Feature/sqlite wal

shushinde:
## Types of changes

## Checklist:


ishwetansh:
Went through the PR in good detail. Overall the direction looks solid to me — moving to a single persistent connection per queue and making WAL opt-in per label is a clean approach, and the test coverage around the WAL behaviour (sidecar files, persistence across reconnect, and the rollback -> WAL migration test) is quite thorough. Nice work on that part.

Since this is touching the data queue persistence which is a fairly critical piece, I have a few concerns I would like us to close out before merging — mostly around the failure paths of the new persistent connection and a couple of silent-failure spots, plus one small doc nit. Nothing that changes the overall approach. A couple of them (specially the connection leak on the init failure path) I would really like to see addressed before this goes in.

Will add the comments inline one by one.

AEPServices/Sources/dataqueue/SQLiteDataQueue.swift
```diff
@@ -35,19 +37,45 @@ class SQLiteDataQueue: DataQueue {
...
+        if config.journalMode == .wal, !SQLiteWrapper.enableWAL(database: connection) {
+            Log.warning(
+                label: LOG_PREFIX,
+                "Failed to enable WAL mode for database '\(databaseName)'."
+            )
+        }
+
         guard createTableIfNotExists(tableName: SQLiteDataQueue.TABLE_NAME) else {
             Log.warning(label: LOG_PREFIX, "Failed to initialize SQLiteDataQueue with database name '\(databaseName)'.")
             return nil
```
ishwetansh, line 56: One thing I want us to fix here — if `createTableIfNotExists` fails we return nil, but by this point we have already opened the connection above and stored it in `self.connection`. Since this is a failable init returning nil, the object is never fully initialised, so `deinit` will NOT be called, and `connection` is an `OpaquePointer` so ARC won't clean it up either. So the sqlite3 handle and its fd end up leaking on this path.

Earlier this was not an issue because `createTableIfNotExists` used to open and close its own connection with a `defer`. Can we `disconnect(database: connection)` before returning nil here (and same for the guard at the open, if we ever add code after it)? This will mostly show up in low-storage / corrupt-db situations, which is exactly when we don't want to be leaking fds.

AEPServices/Sources/dataqueue/SQLiteDataQueue.swift
```diff
@@ -35,19 +37,45 @@ class SQLiteDataQueue: DataQueue {
...
         self.databaseName = databaseName
         self.databaseFilePath = databaseFilePath
         self.serialQueue = serialQueue
+        self.config = config
+
+        guard let connection = connect() else {
+            return nil
+        }
+        
+        self.connection = connection
```
ishwetansh, line 50: Want to flag a behaviour change that comes with holding a single connection for the lifetime of the queue. Earlier every op did `connect()` -> `sqlite3_open(create: true)` -> `disconnect`, so if iOS purged the db file from `.cachesDirectory` while the app was still running, the next op would just recreate the file and writes kept working — the queue kind of self-healed.

Now we open once and keep the handle around. If the OS purges the file underneath us mid-session, sqlite keeps writing to the unlinked inode and that data silently disappears on the next launch, with no error surfaced anywhere. `.cachesDirectory` is purgeable so this is a real scenario (not very frequent, but real). Did we consider a lazy reconnect / reopen-on-write-failure so we keep the self-healing behaviour? Not blocking, but since this is event data I wanted to make sure it is a conscious decision.

AEPServices/Sources/dataqueue/SQLiteWrapper.swift
```diff
@@ -159,4 +159,31 @@ public struct SQLiteWrapper {
...
+    /// - Returns: True if WAL journal mode was successfully enabled, otherwise false
+    @discardableResult
+    public static func enableWAL(database: OpaquePointer) -> Bool {
+        guard let result = query(database: database, sql: "PRAGMA journal_mode=WAL;"),
+              let mode = result.first?.values.first, mode.lowercased() == "wal" else {
+            Log.warning(label: LOG_PREFIX, "Failed to enable WAL journal mode for database.")
+            return false
+        }
+
+        if !execute(database: database, sql: "PRAGMA synchronous=NORMAL;") {
```
ishwetansh, line 185: `synchronous=NORMAL` together with WAL means a committed transaction can be lost on a power-loss / hard OS crash (anything since the last checkpoint), whereas the old rollback + FULL fsync'd on every commit. So we are effectively trading a bit of crash durability of already-committed entities for the I/O win.

It's opt-in so most likely fine, but can we call this out explicitly in the doc comment (and wherever `.wal` gets chosen)? So that a consumer picking WAL knows they are also accepting NORMAL durability and not just the perf benefit. Right now the two are bundled together with no way to keep FULL if someone wanted WAL + full durability.

AEPServices/Sources/dataqueue/DataQueueService.swift
```diff
@@ -28,14 +28,18 @@ public class DataQueueService: DataQueuing {
     public init() {}
 
     public func getDataQueue(label databaseName: String) -> DataQueue? {
+        return getDataQueue(label: databaseName, config: DataQueueConfig())
+    }
+
+    public func getDataQueue(label databaseName: String, config: DataQueueConfig) -> DataQueue? {
         storeQueue.sync {
             if let queue = store[databaseName] {
```
ishwetansh, line 36: This ordering is slightly fragile — the config is applied only at creation and then the queue is cached forever (the comment here says as much). Since `DataQueueService` is shared, if some other module happens to call the plain `getDataQueue(label:)` first, the queue gets built with the default config, and a later `getDataQueue(label:, config: .wal)` for the same label will just return the already-cached default queue and silently ignore the passed config. So the WAL tuning gets dropped with nothing in the logs.

Can we at least log a warning when `getDataQueue(label:, config:)` is called with a non-default config for a label that is already present in `store`? That way if the ordering contract does get broken, it is at least visible instead of a silent no-op.

AEPServices/Sources/dataqueue/SQLiteDataQueue.swift
```diff
@@ -177,6 +186,10 @@ class SQLiteDataQueue: DataQueue {
     func close() {
         serialQueue.sync {
             isClosed = true
+            if let connection = connection {
+                disconnect(database: connection)
```
ishwetansh, line 187: Small thing — `disconnect()` returns a `Bool` (sqlite3_close can come back non-OK) but we ignore it and set `connection = nil` regardless. So if close ever fails we drop the handle with no way to retry it. In practice we finalise our statements via exec/query so a failing close is unlikely here, so low priority — but might be worth at least logging the `false` case so we're not blind to it.

AEPServices/Sources/dataqueue/SQLiteDataQueue.swift
```diff
@@ -35,19 +37,45 @@ class SQLiteDataQueue: DataQueue {
...
+
         guard createTableIfNotExists(tableName: SQLiteDataQueue.TABLE_NAME) else {
             Log.warning(label: LOG_PREFIX, "Failed to initialize SQLiteDataQueue with database name '\(databaseName)'.")
             return nil
         }
     }
 
+    deinit {
+        if let connection = connection {
+            disconnect(database: connection)
```
ishwetansh: Consistency point — every other place touches `connection` / `isClosed` inside `serialQueue.sync`, but `deinit` calls `disconnect` outside it. Practically `deinit` only fires when the refcount hits 0 (and `store` retains the queues for the whole lifetime, so realistically this is only tests / `ServiceProvider.reset`), so I don't think it can actually race today. But just to keep the same serial-queue discipline everywhere, it might be cleaner to route this through the serial queue as well. Non-blocking.

AEPServices/Sources/dataqueue/DataQueueConfig.swift
```diff
@@ -0,0 +1,24 @@
...
+ of the License at http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software distributed under
+ the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ OF ANY KIND, either express or implied. See the License for the specific language
+ governing permissions and limitations under the License.
+ */
+
+import Foundation
+/// Tuning a consumer registers for its `DataQueue` ahead of creating it.
```
ishwetansh: Tiny nit — this first line does not quite read as a proper sentence ("Tuning a consumer registers for its `DataQueue` ahead of creating it."). Since it's public API doc it will show up like this in the generated docs. Maybe reword to something like "Configuration a consumer registers for a `DataQueue` before it is created." Also there is a missing blank line between the `import Foundation` and this doc comment.

shushinde: fixed

AEPServices/Sources/dataqueue/SQLiteDataQueue.swift
shushinde: yes 
wraped inside serialQueue.sync

AEPServices/Sources/dataqueue/DataQueueService.swift
shushinde: yes applied the suggestions

AEPServices/Sources/dataqueue/SQLiteDataQueue.swift
shushinde: added a self healing mechanism and handled the case where dir is purged

AEPServices/Sources/dataqueue/SQLiteDataQueue.swift
shushinde: applied

AEPServices/Sources/dataqueue/SQLiteDataQueue.swift
shushinde: inside disconnect method of sqlitedataqueue class i've added this suggestion

AEPServices/Sources/dataqueue/SQLiteWrapper.swift
shushinde: The consumer is extensions such as edge 
so it is not opt in

AEPServices/Sources/dataqueue/DataQueuing.swift
```diff
@@ -14,8 +14,15 @@ import Foundation
...
     /// - Parameter label: the label you assigned to the `DataQueue` at creation time.
     /// - Returns: the object of `DataQueue`, return false if failed to create an object
     func getDataQueue(label: String) -> DataQueue?
+
+    /// Initialize a `DataQueue` object with the given configuration, applied when the queue is created.
+    /// - Parameters:
+    ///   - label: the label you assigned to the `DataQueue` at creation time.
+    ///   - config: tuning for the queue's database.
+    /// - Returns: the object of `DataQueue`, or nil if it failed to create an object.
+    func getDataQueue(label: String, config: DataQueueConfig) -> DataQueue?
```
akhiljain1907, line 27: Since we are not adding a default implementation for this new api, so our existing extensions(Campaign and Analytics) which uses Mock Utils will break as they need to conform to this protocol. We should use Core's own Mock util there as well like we do in Edge. Let's create a Jira for this and update the Code in those extension.

akhiljain1907:
PR looks good, Approved.
As we discussed in case when a device is locked while db is being written, we can have file protection being set to "NSFileProtectionCompleteUntilFirstUserAuthentication" which is a separate enhancement. Let's also create a separate Jira for this.

akhiljain1907 (approved):

ishwetansh (approved):

